PR #14, opened 2012-11-23
Constructor & warning fix

arkusuma:
I'm working on GPS/GPRS tracking system based on Arduino, and your MQTT library is really helpful, thanks. Here's some fix I caught on the way:

* Domain is not initilized in constructor.  This will cause problem when creating PubSubClient at runtime using new. connect() wrongly detect my IP server as domain name.
* On arduino 1.0.2, publish_P() generate warnings from compiler about unused attribute, and signed & unsigned type mismatch
* PubSubClient(Client& client); doesn't seems to be useful? Because you can't set other mandatory parameters after construction (no setters, private fields).

### DIFF
--- a/PubSubClient/PubSubClient.cpp
+++ b/PubSubClient/PubSubClient.cpp
@@ -7,15 +7,12 @@
 #include "PubSubClient.h"
 #include <string.h>
 
-PubSubClient::PubSubClient(Client& client) {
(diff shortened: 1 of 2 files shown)


knolleary:
The client has updated considerably since this PR and I think all of your original issues are now resolved.

PR #44, opened 2014-10-01
PlatformIO Library Registry manifest file

ivankravets:
* This library in Web: http://platformio.org/#!/lib/show/89/Arduino-PubSubClient
* Specification: [PlatformIO Library Manager](http://docs.platformio.ikravets.com/en/latest/librarymanager/index.html)
* Integration: [IDE Integration](http://docs.platformio.ikravets.com/en/latest/ide.html)

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,12 +1,14 @@
 {
-  "name": "Arduino-PubSubClient",
+  "name": "PubSubClient",
   "keywords": "ethernet, mqtt, m2m",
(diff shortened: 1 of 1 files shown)


skorokithakis:
Does merging this mean that PlatformIO will update the library to the latest master? That'd be great.

knolleary:
I have no idea - @ivankravets, I assume by accepting this file, I'll also need to keep it in sync with the `library.properties` file when I cut new releases?

PR #45, opened 2014-10-08
set char* to const to ease String usage

mgth:
hi,
char * parameters should be set const to allow passing String.c_str() without ugly casts or memcopy.
As pubsubclient never alter char* contents I see no side effect.

I think callback should have been set const too, but it would compromise backward compatibility.

Thank you for this very useful lib.

Mgth


knolleary:
done in 2.0 release

Issue #46, opened 2014-11-07
PubSubClient.h:72: error: section attribute not allowed for ‘<anonymous>’


ntruchsess:
when compiling with Arduino IDE 1.0.6 you get the error:
PubSubClient.h:72: error: section attribute not allowed for ‘<anonymous>’

with Arduino IDE 1.0.5 it did compile just fine.

removing the 'PROGMEM' from this line solves the issue.
https://github.com/knolleary/pubsubclient/blob/master/PubSubClient/PubSubClient.h#L72

jasonmhite:
Same error on 1.6.4...

knolleary:
Bit long ago now, but what board were you targeting? It compiles fine for me against a regular arduino board definition.

knolleary:
Fixed in the newly released 2.0 version

niwantha33:
install this version 
https://github.com/heman4t/Arduino-pubsubclient

PR #47, opened 2014-11-07
PubSubClient.h fix error: section attribute not allowed for <anonymous>

ntruchsess:
fix for #46


knolleary:
Fixed in the 2.0 release

Issue #51, opened 2014-12-18
Question: How to change the port settings dynamically


marcelrv:
Is there a way to set the port dynamically, meaning not directly at the initialization.
I am reading the server and the port int from eeprom, however then I'm no longer able to set the port settings. For the mqtt server it is still possible to change this later, but for the port not.

Is there another way other then introducing something like
void PubSubClient::setPort(uint16_t port) {
	this->port = port;
}

knolleary:
Appropriate setter added

PR #52, opened 2014-12-18
Allow for setting of the port after first initiation

marcelrv:
as per question #51


knolleary:
Fixed by #60

marcelrv:
cool.... will use the new version soon.
Thanks for the library!

PR #57, opened 2015-02-28
Intel Galileo and Edison Support

anupam19:
added #include <avr/pgmspace.h> and this library is now compatible with Intel Galileo and Intel Edison boards


knolleary:
Have approached this slightly differently with acc9fe500e4cac80707e526b95f4fdcddb9082c5 - put `ifdef PROGMEM` around references to progmem sections. That means we can support more platforms that don't have progmem, nor have avr/pgmspace.h available.

Issue #58, opened 2015-03-10
Request: set hostname in connect method rather than constructor?


davidmonro:
I've written some code which stores the DNS name in EEPROM, and then copies it from the EEPROM to RAM at runtime (so I don't have to recompile to change the hostname, just reflash the EEPROM). However the PubSubClient expects the hostname in the constructor, and I don't think (correct me if I am wrong) that arduino supports dynamic instantiation using new()?

Currently passing the constructor a pointer to the array that I fill in later from the EEPROM works, but only because the class just maintains a pointer to the array I've passed it; if it happened to make a copy of it internally my code wouldn't work. And relying on the internal implementation like this is horribly inelegant :)

Would it be possible to add a constructor variant which doesn't take the hostname and port as arguments, and then either provide a variant of the connect() method which takes those arguments, or alternatively a separate method to set them?

(I also store the ClientID in the EEPROM, but that is supplied to the connect() method and so doesn't suffer from the same issue).

Cheers

David

knolleary:
Added a setServer function that allows you to set the domain/IP/port after the constructor is called.

New release coming this week (finally!)

PR #60, opened 2015-04-08
refactored constructors to use getters and setters

vicatcu:
I believe this addresses Issue #58 (at least). I implemented some setters that allow for dynamic runtime configuration of the PubSubClient. The most likely use case I can think of is using the empty constructor to create a global instance, then in setup() assigning the various fields based on configuration. I also don't believe this pull request alters the existing examples and usage patterns.

### DIFF
--- a/PubSubClient/PubSubClient.cpp
+++ b/PubSubClient/PubSubClient.cpp
@@ -8,42 +8,46 @@
 #include <string.h>
 
 PubSubClient::PubSubClient() {
-   this->_client = NULL;
-   this->stream = NULL;
+   this->_client = 0;
+   setCallback(NULL);
+   setBrokerIP(ip);
+   setPort(1883);
+   setBrokerDomain(NULL);
+   this->stream = 0;
 }
 
 PubSubClient::PubSubClient(uint8_t *ip, uint16_t port, void (*callback)(char*,uint8_t*,unsigned int), Client& client) {
-   this->_client = &client;
-   this->callback = callback;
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for this - I don't get much time to play with the library, so I appreciate the contribution.

It all looks sensible to me - as you say, they absolute key thing is that it's backwards compatible. If you think its good on that front, then I'm happy to merge.

vicatcu:
@knolleary it's my pleasure to contribute! I know a lot of people use this library, myself included. I'm using the pull requested version right now and it certainly works. I would really love to see at least one other user do some regression on the PR besides myself (preferably more), but it is pretty straightforward.

vicatcu:
@knolleary absent comments from other people, I've been using the methods in this pull request on the new Air Quality Egg v2 baseline software for a while and it seems to work quite well. I'd like to see this PR merged, what do you think?

vicatcu:
thanks!

Victor Aprea // Wicked Device
victor.aprea@wickeddevice.com
T: @vicatcu <http://twitter.com/vicatcu/>

knolleary:
Took a while... but I got there ;)

knolleary:
I've changed the gettings/setters slightly from what was in this PR - rather than individual ones for ip/domain/port, I've done:

       void setServer(uint8_t * ip, uint16_t port);
       void setServer(char * domain, uint16_t port);

Bit more streamlined - unlikely you'd want to change one but not the other.

PR #65, opened 2015-06-19
Fixed  "char *" depreciation error

stuthedew:
Fixed  "char *" depreciation error by changing to "const char*"


knolleary:
done in the 2.0 release

PR #68, opened 2015-08-26
Removed "PROGMEM" from publish_p method, as it won't compile otherwise.

l-sch:
The library won't compile in Arduino without "PROGMEM" removed.

### DIFF
--- a/PubSubClient/PubSubClient.cpp
+++ b/PubSubClient/PubSubClient.cpp
@@ -281,7 +281,7 @@ boolean PubSubClient::publish(char* topic, uint8_t* payload, unsigned int plengt
    return false;
 }
 
-boolean PubSubClient::publish_P(char* topic, uint8_t* payload, unsigned int plength, boolean retained) {
+boolean PubSubClient::publish_P(char* topic, PGM_P payload, unsigned int plength, boolean retained) {
    uint8_t llen = 0;
    uint8_t digit;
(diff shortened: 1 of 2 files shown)


knolleary:
I'm out of touch with the latest arduino ide - but simply removing PROGMEM is not the right fix here as it will break anyone making use of the method. Need to understand where PROGMEM went and how it should be done today.

l-sch:
Ok, so I'm not all into that PROGMEM-thingie and just wanted your, by the way awesome, library to compile.

But maybe we could solve this issue if "const uint8_t PROGMEM *" is replaced by "PGM_P" as I did in my second commit. Could you test if this works?

knolleary:
I've just installed the latest Arduino IDE - 1.6.5, and it compiles the library fine as-is.

Can you provide some more information about what environment you're trying to compile it in?

l-sch:
So I actually just figured out what the real problem is.

I'm trying to compile you're library for an ESP8266-Chip, which isn't even ARM and therefor doesn't support something like *PROGMEM*. And thats, why it can't be compiled for the ESP.

Huge thanks anyway!

knolleary:
Ahh - that explains it. Now, we could ifdef around the PROGMEM stuff so it only gets included if it's available...

l-sch:
Yeah. Will fork again and start a new Pull-Request.

PR #69, opened 2015-08-26
Made Library compile on ESP8266-Modules.

l-sch:
On an ESP8266-Chip, the *PROGMEM*-based publish-method doesn't get compiled, as it can't be used anyways.


knolleary:
Been looking at this some more - it isn't just the ESP8266 platform that's broken by PROGMEM - the Intel Galileo also doesn't have it (for example).

So, rather than ifdef using the ESP8266 symbol, it would be better to ifdef on PROGMEM itself - that will then catch all platforms. I'm in the middle of some changes here, so will apply it myself. But thank you for raising this and bringing it to my attention.

PR #73, opened 2015-09-01
Changed "or" to "and" to fix issue with publishing only implementations.

CplSyx:
When testing this with a device that only publishes, the connection eventually times out and then needs to be re-established for EVERY message published. 
Changing this fixes the problem, and does not appear to affect the normal operation.

### DIFF
--- a/PubSubClient/src/PubSubClient.cpp
+++ b/PubSubClient/src/PubSubClient.cpp
@@ -263,7 +263,7 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
 boolean PubSubClient::loop() {
     if (connected()) {
         unsigned long t = millis();
-        if ((t - lastInActivity > MQTT_KEEPALIVE*1000UL) || (t - lastOutActivity > MQTT_KEEPALIVE*1000UL)) {
+        if ((t - lastInActivity > MQTT_KEEPALIVE*1000UL) && (t - lastOutActivity > MQTT_KEEPALIVE*1000UL)) {
             if (pingOutstanding) {
                 this->_state = MQTT_CONNECTION_TIMEOUT;
                 _client->stop();


knolleary:
Odd, I have a specific test for that very scenario, which is failing with your change - https://travis-ci.org/knolleary/pubsubclient/builds/78299217#L148

I don't have an arduino to hand at the moment - do you have a simple test sketch that reproduces it?

CplSyx:
I've taken the relevant sections from my app and put them into the Basic MQTT example to keep it as simple as possible. The following code produces the error.

```
#define UUID "66b604c0445211e5b9700800200c9a66"

#include <SPI.h>
#include <UIPEthernet.h>
#include <PubSubClient.h>

byte mac[]    = {  0xDE, 0xED, 0xBA, 0xFE, 0xFE, 0xED };
IPAddress ip(192, 168, 1, 44);
IPAddress server(192, 168, 1, 23);

unsigned long ticker;

bool tickedOver()
{
  if(((signed long)(millis()-ticker))>0)
    return true;
  else
   return false; 
}
void tick()
{
  ticker = millis()+5000;
}

void callback(char* topic, byte* payload, unsigned int length) {
  Serial1.print("Message arrived [");
  Serial1.print(topic);
  Serial1.print("] ");
  for (int i=0;i<length;i++) {
    Serial1.print((char)payload[i]);
  }
  Serial1.println();
}

EthernetClient ethClient;
PubSubClient client(ethClient);

void reconnect() {
  // Loop until we're reconnected
  while (!client.connected()) {
    Serial1.print("Attempting MQTT connection...");
    // Attempt to connect
    if (client.connect("arduinoClient", "device", "device")) {
      Serial1.println("connected");
      // Once connected, publish an announcement...
      client.publish("outTopic","hello world");
      // ... and resubscribe
      client.subscribe("inTopic");
    } else {
      Serial1.print("failed, rc=");
      Serial1.print(client.state());
      Serial1.println(" try again in 5 seconds");
      // Wait 5 seconds before retrying
      delay(5000);
    }
  }
}

void sendHeartbeat()
{
  String heartbeatString =  "{\"uuid\":\"";
  heartbeatString += UUID;
  heartbeatString += "\",\"heartbeat\":{\"time\":\"";
  heartbeatString += millis();
  heartbeatString += "\"}}";
  client.publish("server.heartbeat.66b604c0445211e5b9700800200c9a66.info.test", heartbeatString.c_str());
  Serial1.println("Sent Heartbeat" + heartbeatString);
}

void setup()
{
  Serial1.begin(9600);

  client.setServer(server, 1883);
  client.setCallback(callback);

  Ethernet.begin(mac, ip);
  // Allow the hardware to sort itself out
  delay(1500);
}

void loop()
{
  if (!client.connected()) {
    reconnect();
  }
  client.loop();
  if(tickedOver())
  {
    sendHeartbeat();
    tick();
  }
}
```

This appears as in the console monitor as 
```
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"3084"}}
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"8118"}}
[...normal sending for some time...]
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"98582"}}
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"103608"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"108636"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"113664"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"118691"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"123719"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"128747"}}
Attempting MQTT connection...connected
Sent Heartbeat{"uuid":"66b604c0445211e5b9700800200c9a66","heartbeat":{"time":"133775"}}
Attempting MQTT connection...connected
```

Can be seen in RabbitMQ admin page as new connections being created.

knolleary:
Thanks for sharing the code.

Your topic and payload are ~130 characters long. The default maximum message size is 128 bytes. This is causing a buffer overrun in the client and overwriting the value of `lastOutActivity` which causes the keepalive logic to fail.

If you want to publish using these topics/payloads, you'll need to increase the value of `MQTT_MAX_PACKET_SIZE` in `PubSubClient.h`. I increased it to 256 and your sketch runs fine.

CplSyx:
Nick - thank you. How I didn't spot this I have no idea... really appreciate the callout on the max packet size. 
Feel a bit stupid for wasting a fair amount of time debugging something that wasn't even an issue in the first place, but you live and learn eh!

knolleary:
Heh - no problem. It made me realise the library could do a better job of protected itself - so the subscribe/publish functions now reject calls where the topic/payload are too big, rather than try to handle them and fail in unexpected ways. So for that, thank you!

PR #75, opened 2015-09-09
Example for ESP8266

marcelrv:
I was very happy to see this working well. Now ESP8266 and MQTT becomes even more easy.
Maybe attached example will help others to quickly get started

### DIFF
--- a/examples/esp8266/mqtt_esp8266.ino
+++ b/examples/esp8266/mqtt_esp8266.ino
@@ -1,0 +1,124 @@
+/*
+ Basic ESP8266 MQTT example
+
+ This sketch demonstrates the capabilities of the pubsub library in combination with the ESP8266 board/library.
+ It connects to an MQTT server then:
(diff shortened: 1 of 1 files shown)


knolleary:
Perfect - thanks!

anupam19:
Are  you using Lua Firmware or the AT based libraries?

marcelrv:
I made my testing on the nodemcu 0.9 board, I think I last loaded it with the lua firmware

anupam19:
Thanks for the info. So this #include <ESP8266WiFi.h> works ONLY with Lua Firmware. Right? I am asking because I work with AT commands

marcelrv:
I haven't tested it out yet on one of my other modules, but my current understanding is that it does not matter if it is based on lua or AT firmware.

anupam19:
Haven't seen any AT based command in the library. That's why I doubt it is only for nodemcu

PR #80, opened 2015-09-28
only define max msg len if not defined yet

edelache:
Allows this setting to be redefined without modifying the pubsubclient.h file.

### DIFF
--- a/src/PubSubClient.h
+++ b/src/PubSubClient.h
@@ -20,7 +20,9 @@
 #define MQTT_VERSION MQTT_VERSION_3_1_1
 
 // MQTT_MAX_PACKET_SIZE : Maximum packet size
+#ifndef MQTT_MAX_PACKET_SIZE
 #define MQTT_MAX_PACKET_SIZE 128
+#endif
 
 // MQTT_KEEPALIVE : keepAlive interval in Seconds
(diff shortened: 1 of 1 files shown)


knolleary:
Hi, thanks for this.

I remember when I first wrote the library I couldn't get this to work in a meaningful way -  how a user would structure their sketch to define this value without having to edit the library. It isn't a topic I've revisited since then.

Can you describe a bit about how you'd see this being used in practise?

There are a few other #defines in there that we'd want to expose this way.

edelache:
I just use a #define in my main file before including PubSubClient.h, so
that when PubSubClient hits its code, its already defined, and therefore
does not change it.

Eric

knolleary:
Have you verified that actually works? I tried exactly that at the time, but due to the way the ide pulled out #includes in its preprocessor, you couldn't have a define before an include in your sketch.

If that works for you, then clearly the preprocessor has changed and life gets easier for users who want to modify these defines.

edelache:
Hmm, yes it does work for me, I'm not using arduino, but esp8266. Same IDE
though. I'll double check again tomorrow and let you know if its working.

Eric

edelache:
You know, I thought it was working but I think you're right. It seems the
preprocessor is not picking up the define in the right order.

Eric

PR #96, opened 2015-12-04
correct handling of dns fauilure

ElvisTheKing:
In case of domain name resolution error result can be negative (see Dns.cpp:46)


Suxsem:
this should definitely be merged

Issue #59, opened 2015-04-08
Add support for QOS1


CptanPanic:
For some use cases such as a state transition, it is required to make sure message was delivered.  I understand your concerns about local arduino memory, but how about an option?

ccfiel:
@knolleary Hello great project! :) I just wondering there is QOS 1 in subscribing. Why there is no QOS 1 publishing?

knolleary:
Because there are no persistence requirements to receive a qos1 message,
but there are for sending.

ccfiel:
@knolleary Example I have a sensor and publish the reading every 10 mins and It would be publish just QOS 0. The problem is the one subscribing the reading and doing computation of the sensor would possibly cannot get all the sensor reading because its was just publish with QOS 0. Would that be correct? or did i miss something?

intel2004:
QOS1 for publishing is useful in some case, i know that arduino uno is not enoght capable to do that, but a Mega 2560 board?
ethernet shield example use less than 10 percent of the board!

rhbroberg:
I'd like to resurrect this conversation: the use case I'd like to be able to fulfill is an arduino using this library be able to send a message which can be durable for the subscriber.  Message durability in MQTT only works if the publisher uses at least QoS > 1, and the subscriber sets clean session to false.  If my MQTT server is up, but the service which has subscribed to those topics is down, I lose all of the information which arrives in the interim.  I'd like to be able to leverage durability and have my subscriber decide if it wants to deal with messages which arrived while it was offline.  This subscriber can be implemented in another language on another platform which has plenty of resources.  Most (all?) of the client libraries default to 'clean session' of true, so by default nothing would change for those users either.

If it's a semantics issue, don't call it QoS of 1; call it whatever you want since it's not practical to implement a persistence store on the publisher's side in this library.  Ignore the PUBACK messages, don't promise the Arduino sender it's guaranteed at least once.  But allow the subscriber to be restarted and not miss any data.

I'd be happy to resubmit the pull request with the changes limited in scope and without breaking the existing interface.

Issue #55, opened 2015-02-13
Rapid Messages will Crash the Client?


kerti:
Hi! First of all, kudos on your great work.

I've been using your library to subscribe to an MQTT topic, and then relay the message to another Arduino "field unit" via RF links using [RadioHead](http://airspayce.com/mikem/arduino/RadioHead/). Here's the sketch if you'd like to take look at it:

https://github.com/kerti/catleya-arduino/blob/master/baseStation/baseStation.ino

I use openHAB to publish messages to MQTT based on what switch I press on my handset. So, all goes well when I test one switch and just leave it alone. But when I mash on the controls and make openHAB spit out a bunch of messages rapidly, eventually the Arduino just goes silent, and I presume it's crashed.

I don't know what happened, maybe you could check it out and see if I crashed the client?

knolleary:
Hi, I don't see anything obviously wrong with your code.

I've not generally had an issues like this, but then I'm not sure I've ever tried swamping the device with messages.

What have you done to try to debug it?

kerti:
All I've done is add some code to check my available memory, because my first suspicion was that I somehow used up all memory. This isn't the problem. I haven't tried anything else.

If you have any other ideas I'd be happy to try it out and report back.

kerti:
Reporting back, I think we simply shouldn't send that many messages at a time. I modified my server to limit itself to a maximum of 1 message per 10 seconds, and it seemed to work fine.

swell-d:
I had the same problem. After some time my arduino stops to respond. 
Problem was with my DHCP server. I set static IP on my arduino and my router set the same IP to another device on my network =)

swell-d:
kerti, how did you limit the server?

kerti:
swell-d, I use openHAB to send my MQTT messages. What server do you use?

swell-d:
The same. I use MQTT event bus for outgoing messages.

kerti:
@swell-d, I simply created an item that I don't show on my UI, and use rules to compare it to the actual items in the UI. An example would make explaining easier:

I have an 8-channel relay attached to an Arduino which I complained about at the top of this thread. I currently have these items:

```xtend
gRelays_Garage_RELAY01_Device01
gRelays_Garage_RELAY01_Device02
gRelays_Garage_RELAY01_Device03
gRelays_Garage_RELAY01_Device04
gRelays_Garage_RELAY01_Device05
gRelays_Garage_RELAY01_Device06
[...]

PR #100, opened 2015-12-20
publish qos 0,1,2; loop() before publish and subscribes; others

### LINKED ISSUE #98: subscribe more than 5 topics in sequence
Hi,

at first I like to say thanks for your great work!

Right now I try to subscribe more than 5 topics in sequence.

```
//Loop until we're reconnected
  while (!mqtt_client.connected() && WiFi.status() == WL_CONNECTED) {
       Serial.print("WIFI-Status: ");
       //Serial.println(WiFi.status());
    Serial.print("Attempting MQTT connection...");
    // Attempt to connect
    if (mqtt_client.connect(char_MQTT_DeviceName)) {
      Serial.println("connected");
      //broker_subcribe();

  mqtt_client.subscribe("36/Output/01");
  mqtt_client.subscribe("36/Output/02");
  mqtt_client.subscribe("36/Output/03");
  mqtt_client.subscribe("36/Output/04");
  mqtt_client.subscribe("36/Output/05");
  //mqtt_client.subscribe("36/Output/06");
  //mqtt_client.subscribe("36/Output/07");
  //mqtt_client.subscribe("36/Output/08");
 
``` 
Subscribing 5 topics work fine......
Subscribing more then 5 topics in sequence will force a disconnect from the broker.

I´m not shure where to search the fault:

   - Is thies a PubSubClient problem
   - or a Mosquitto MQTT problem 
   - or, how so often, my problem...... :-)

Can anybody check this out...
Thanks for your Help
Pf@nne

Suxsem:
- default MQTT_KEEPALIVE increased to 30 seconds
- esp8266 example updated with
  1) onDisconnected callback
  2) non blocking reconnect
  3) authentication
  4) last will message
  5) payload cast to string
- loop() now processes ALL available mqtt packets in the tcp buffer, not just the first
- publish(), subscribe() and unsubscribe() calls are preceded by loop() call to process incoming messages (and ACKS) without filling the TCP buffer when multiple pub/sub actions are performed
- publish() now supports QoS 0, 1 and 2. See the updated readme for more info
- improved comments in readPacket(uint8_t\* lengthLength) to better understand what it does

Signed-off-by: Stefano Semeraro semeraro.stefano@hotmail.it

Resolve #98, #59, #55

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,14 @@ Full API documentation is available here: http://pubsubclient.knolleary.net
 
 ## Limitations
 
- - It can only publish QoS 0 messages. It can subscribe at QoS 0 or QoS 1.
+ - It can subscribe at QoS 0 or 1.
+ - It can publish at QoS 0, 1 or 2. WARNING! No retransmission is supported to
+   keep the library as much memory friendly as possible. (Without retransmission
+   support, the publish QoS is only meaningful when the broker sends your
+   message to a subscriber, supposing that the subscriber subscribes with a QoS
+   greater then or equal to the publish QoS; consider that MQTT runs over TCP,
+   so retransmission isn't really required in most cases, especially when
+   publishing to the broker)
  - The maximum message size, including header, is **128 bytes** by default. This
    is configurable via `MQTT_MAX_PACKET_SIZE` in `PubSubClient.h`.
  - The keepalive interval is set to 15 seconds by default. This is configurable
--- a/examples/mqtt_esp8266/mqtt_esp8266.ino
+++ b/examples/mqtt_esp8266/mqtt_esp8266.ino
@@ -11,9 +11,7 @@
   - If the first character of the topic "inTopic" is an 1, switch ON the ESP Led,
     else switch it off
 
- It will reconnect to the server if the connection is lost using a blocking
- reconnect function. See the 'mqtt_reconnect_nonblocking' example for how to
- achieve the same result without blocking the main loop.
+ It will reconnect to the server if the connection is lost
 
  To install the ESP8266 board, (using Arduino 1.6.4+):
   - Add the following 3rd party board manager under "File -> Preferences -> Additional Boards Manager URLs":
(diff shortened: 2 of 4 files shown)


Suxsem:
ops...test suite should be updated to match the new publish() signature :)

knolleary:
Sorry, but I don't accept pull requests that contain multiple changes that have had no discussion. Especially when they are contained in a single commit so it is impossible to distinguish them.

There are also a number of issues with the proposed changes. At a glance, they include:

 - changed default keepalive with no explanation for why
 - change to use a while loop in the loop function is incorrect. In certain circumstances it will mean loop never exits as it is continually processing incoming packets
 - used tabs when the rest of the file uses spaces
 - changed the signature of existing publish functions that will break all existing users - the regression tests are there for a reason.
 - I have explicit not added QoS 2 for a reason


If there are specific changes you want to make, raise an issue so we can discuss.

Suxsem:
I understand. Anyway i made those changes for myself and other people, because me and other students of my course (i'm an engineering student) need those features.

I shared them here because maybe you too find them useful.

If you want to pick a specific change go ahead and make a new commit, i will continue to use my fork.

I want to use this space to thank you for the wounderful library, the most solid mqtt library for esp8266 that I found (now even more solid in my fork :D)

knolleary:
OK, thank you for the contribution, but please take my feedback on board.
It is good practice to have a single feature/fix per commit, especially
when you intend to contribute back to a project - it makes the project
maintainer's task much easier to see what has changed and more importantly,
why it has changed.

When I'm next near my laptop, I will look at the subscribe/loop issue
properly to see what handling is needed, if any, in this code.

Thanks, Nick

Suxsem:
Sure,  I will do that definitely. 
Just two words about the changes. 

1) publish qos. I found in a your comment to an issue that you don't want qos 1 or qos 2 for memory issues (qos 0 doesn't need persistence). But if you have read the Readme changes in my commit you will now that retransmission is not implemented and qos 1 and 2 are only there for the broker - receiver communication. So, if the persistence is not the problem, what is it? 

2) while instead of if in loop. The infinite loop can only happens if the board receives more message in a second than it can process. Considering that 99% of the time this doesn't happens, having the while loop can prevent tcp buffer to fills, causing board to disconnects. Moreover, if the board receives more messages than it can process, we have a bigger problem. So, what circumstances you are referring in loop never ends? 

3) change in signature. We should add a new signature without removing the old one. Anyway this seems to me useless. You could add a warning in the changelog, pushing users to add qos=0 when they set retain to true. They will suddenly read the changelog when their code will not build. 

4) keep alive was changed because all other libraries that I found (and other examples of mqtt clients like Java paho) have their default values far bigger then 30 seconds. Anyway it should be made user defined. What do you think? 

5) tabs were a mistake,  sorry :)

knolleary:
1. not sure what you mean by 'qos1/2 are only there for the broker-receiver communication'. The qos flows are just as important in either direction. But as I think about it some more, the only scenario where persistence is required is if the client supports non-clean sessions - as it would have to retry outstanding q1/2 publishes after a reconnect. But as this client only supports clean session connections then it is feasible to added q1/2 publish.

2. we had a recent issue where someone hit exactly this scenario after modifying the code as you had. It isn't that unlikely - you just need a fairly steady stream of incoming messages. There may be some trade off to be had - let client.loop handle more than one inbound message, but only to some limit, for example 5.

3. I take backwards compatibility very seriously. If it does have to break something, then the major version number needs to be increased to stick by the rules of semantic versioning. But I'm not convinced it does have to break the existing signatures.

Suxsem:
1) let's start with the fact that retransission on esp8266 is not possible due to memory problems. So the esp8266 will always publish to the mqtt broker with no deliver guarantee; now, when the esp8266 send a message, it first deliver it to the broker, then the broker deliver it to the subscriber. let's focus on the delivering of that message to the subscriber (from the broker). The qos of that process it's the maximum between message qos and subscribe qos.
So, if the subscriber (for example my smartphone) subscribes with qos=2, but the esp8266 sends it with qos=0, not only the deliver to the broker has qos=0, but ALSO the deliver from the broker to the subscriber.
My english it's not good, I hope you can understand me :D

2) i can't understand how avoiding the while loop have fixed the problem of this guy. if the while loop never exited, that's mean that the incoming stream it's so quick that the esp can't process it. The only situation that could cause problems is when the mqtt messages are so big that the loop() finds the first bytes but then it must wait for the message to be entirely received.
I think that we could edit the code in this way: only loops the wificlient.available() until the mqtt header is received. then check if the wificlient.available() is equals to the remaining length found inside the header. if not, exit the client.loop() and check again in the next client.loop(), if yes then the message is already entirely in the tcp buffer and can be process without wasting time!

3) i'm sure you have more experience then me in those things :)

MSERTER:
hi. what is the version of the pubsubclient library that supports all QoS qualities? where can I get to this library?

FDR-96 (approved):

Issue #117, opened 2016-01-25
Callback signature


tomkcook:
Is there some particular reason not to use C++11's `std::function<void(char*, uint8_t*, uint32_t)>` as the callback type instead of a bare function pointer?  That would then allow passing pointer to member, lambda, functors etc instead of just ordinary functions.

If there's no objection, I'll create a pull request.

knolleary:
I'll refer you to this comment I made over the weekend: https://github.com/knolleary/pubsubclient/issues/115#issuecomment-174217440

As long as this is fully backwards compatible, so any existing sketch will still work, then go ahead.

tomkcook:
I'm not sure if all the arduino toolchains have enough C++11 to make this work.  I'll experiment and see.

tomkcook:
On a quick survey, *no* other Arduino board supports `std::function<>` (or indeed the `<functional>` header at all).  I'll submit a pull request that makes this change ESP8266-specific, but see what you think.

marcoschwartz:
Hello, since the last update of the library I am having the following error in my library (aREST) when used with the ESP8266:

/Users/marco/Dropbox/Arduino/libraries/aREST/aREST.h:1051: undefined reference to `PubSubClient::setCallback(std::function<void (char*, unsigned char*, unsigned int)>)'

With Arduino-based sketches, no problems. How to fix this? I think it is linked to this issue. Thanks!

knolleary:
@marcoschwartz hmmm, ok. Will need input from @tomkcook on this as this was his change. We can rollback if it has regressed behaviour. Do you have a pointer to your code? I see the aREST library on github doesn't reference setCallback anywhere so I'm guessing this is a development version?

marcoschwartz:
Thanks for the fast answer! Indeed it's not in the library, but I do a client.setCallback(callback); in all the code using PubSubClient.

tomkcook:
I'll have a look.

tomkcook:
@marcoschwartz I can't see the problem.  Can you double-check what you have on line 573 of PubSubClient.cpp?  ie that you have 21b75a2c4aad100a3fc2cb1f04e6a9b147d69a3d in your source tree.

marcoschwartz:
Thanks for the support! Actually I had to revert to 2.4.0 and then update again from the Arduino library manager, and now it works!

Issue #115, opened 2016-01-23
mycustom library and pubsubclient call back


vickeyhort:
Hi. I am trying to make my own library for my Esp8266 Arduino using ESP8266 library and PubSubClient library. Problem is pubsubclient call back function, It is giving error. May someone please help me? Here are ino, h, cpp and error log files.
[h.txt](https://github.com/knolleary/pubsubclient/files/101987/h.txt)
[ino.txt](https://github.com/knolleary/pubsubclient/files/101988/ino.txt)
[Error.txt](https://github.com/knolleary/pubsubclient/files/101989/Error.txt)
[cpp.txt](https://github.com/knolleary/pubsubclient/files/101990/cpp.txt)

knolleary:
Hi,

whilst I can recreate your error, my C++ knowledge is not good enough I know what the fix is.

The problem is this library expects the callback function signature to be:

    void (*callback)(char*,uint8_t*,unsigned int)

but you are providing a callback that is a member of a class:

    void EspMQTT::callback(char* topic, byte* payload, unsigned int length)

I don't know how to make those two match up in C++.

tomkcook:
@vickeyhort - see the repository tomkcook/pubsubclient for a version that can handle this.  You can set a callback from a class member function like this:

```
mqtt.set_callback([this] (char* topic, byte* payload, unsigned int length) { this->callback(topic, payload, length); }
```

marvinroger:
@tomkcook the @knolleary "version" can also handle this with the ESP8266, as the OP requested:

```c++
mqtt.setCallback(std::bind(&YourClass:_yourCallback, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
```

@vickeyhort, can you close this issue?

vickeyhort:
@tomkcook Sorry for such late reply, I was busy with one of my project. As you mentioned;
`mqtt.set_callback([this] (char* topic, byte* payload, unsigned int length) { this->callback(topic, payload, length); }`
Where should I use this piece of code? In my cpp file or ino file? I am really not understanding it.

jon-nfc:
> @tomkcook the @knolleary "version" can also handle this with the ESP8266, as the OP requested:
> 
> ```c++
> mqtt.setCallback(std::bind(&YourClass:_yourCallback, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3));
> ```
> 
> @vickeyhort, can you close this issue?

[...]

PR #120, opened 2016-01-27
Use std::function on ESP8266 platform.

tomkcook:
As discussed in #117.  For non-ESP8266 platforms, `<functional>` is not available; for these platforms, the only change is to use MQTT_CALLBACK_SIGNATURE in the definition of `PubSubClient::setCallback`, which seems like it was just an oversight anyway.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -570,7 +570,7 @@ PubSubClient& PubSubClient::setServer(const char * domain, uint16_t port) {
     return *this;
 }
 
-PubSubClient& PubSubClient::setCallback(void(*callback)(char*,uint8_t*,unsigned int)){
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for raise this. Just a heads up I'm tied up for the next few days with work - won't get a chance to look at this before then. So, not ignoring it... just not enough hours in the day this week.

tomkcook:
Apologies - that works under ESP8266 but I guess not on other toolchains.  Lack of C++11 support in the Arduino toolchains is a bit frustrating!

Pfannex:
Hi,

can someone post how to set the callback from my own Class outside the main Sketch?

How to declarate the callback function in my own Class?
How to set the callback inside the konstructor?

Pfannex:
Hi,

found it in #115

PR #123, opened 2016-01-29
Added an API in order to auto-negotiate protocol 3.1  vs 3.1.1 at runtime

mirkoflchtt:
Hello, 
I integrated this library in a domotic project that uses Domoticz.
I had a problem interfacing Mosquitto Broker on Ubuntu 12.04 simply because there the protocol 
was "old" 3.1 and the pubsubclient library instead was statically compiled to 3.1.1.
Since I noticed that with few changes it was feasible to switch the protocol at runtime, I thought that
it make sense to add a simple API to set the protocol and a simple kind of auto-negotiation (i.e. auto ping-pong protocol switching when a timeout is reached when trying to connect to the broker)
  
I checked the code was working as expected by setting initially the protocol to 3.1.1 and connecting
to the broker supporting the 3.1: 1st time the timeout was reached, but at the 2nd trial the library switched to 3.1.1 automatically. 
I think that this is useful to enhance the reliability of the library..

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -5,19 +5,20 @@
 */
 
 #include "PubSubClient.h"
-#include "Arduino.h"
 
 PubSubClient::PubSubClient() {
     this->_state = MQTT_DISCONNECTED;
     this->_client = NULL;
     this->stream = NULL;
     setCallback(NULL);
+    setProtocol(MQTT_VERSION);
 }
 
 PubSubClient::PubSubClient(Client& client) {
     this->_state = MQTT_DISCONNECTED;
     setClient(client);
     this->stream = NULL;
+    setProtocol(MQTT_VERSION);
 }
 
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client) {
@@ -25,12 +26,14 @@ PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client) {
     setServer(addr, port);
     setClient(client);
     this->stream = NULL;
+    setProtocol(MQTT_VERSION);
 }
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client, Stream& stream) {
     this->_state = MQTT_DISCONNECTED;
     setServer(addr,port);
     setClient(client);
     setStream(stream);
+    setProtocol(MQTT_VERSION);
 }
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, MQTT_CALLBACK_SIGNATURE, Client& client) {
     this->_state = MQTT_DISCONNECTED;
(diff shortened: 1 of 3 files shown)


knolleary:
Hi, I prefer issues are raised before PRs so we can discuss the details before energy is invested in writing code.
The reason for #defining it was to minimise the size of the library. So would need to look at what impact that would have.

There are some other code-level issues with the approach and the unit tests haven't been touched, regardless of the above.

src/PubSubClient.cpp
```diff
@@ -128,16 +141,15 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
...
-#endif
-            for (j = 0;j<MQTT_HEADER_VERSION_LENGTH;j++) {
-                buffer[length++] = d[j];
-            }
+			if ( protocolType==MQTT_VERSION_3_1 ) {
+				const uint8_t d[9] = {0x00,0x06,'M','Q','I','s','d','p', MQTT_VERSION_3_1};
+				for ( j=0;j<sizeof(d);j++ )
+                	buffer[length++] = d[j];            
+			} else {
+      			const uint8_t d[7] = {0x00,0x04,'M','Q','T','T',MQTT_VERSION_3_1_1};
```
knolleary: `d` is already defined above, and the for loop is duplicated needlessly

src/PubSubClient.cpp
```diff
@@ -174,12 +186,16 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
...
-
+			
             while (!_client->available()) {
                 unsigned long t = millis();
                 if (t-lastInActivity >= ((int32_t) MQTT_SOCKET_TIMEOUT*1000UL)) {
                     _state = MQTT_CONNECTION_TIMEOUT;
                     _client->stop();
+
+                    // Next time, Try to Switch Protocol if client not responding 
+                    protocolType = (protocolType==MQTT_VERSION_3_1_1) ? MQTT_VERSION_3_1 : MQTT_VERSION_3_1_1;
```
knolleary: Need to put some more thought on if automatically switching is good or not. Typically, you will know (or discover the first time you test) whether the protocol you are using is supported by the broker you are pointed at. You are highly unlikely to need to change protocols at runtime after the initial setup.

mirkoflchtt: Hi,
1st I apologize for not having discussed this earlier..
A you may realize I recently subscibed to github.. hence I'm not yet fully
confident about the way to go to proactively contribute to open-source..

Just few words why I need to implement this: I'm doing a domotic controller
for the blinds in my house.. and you could understand that it is hard to
mantain it once hide in the walls..
So for example I know that in few months I will upgrade my MQTT broker from
12.04 to 16.04 and thus I know that I should eventually recompile
everithing on MCU due to change in protocol..
So i'm trying to compile a FW that still has low memory footprint yet being
a little adaptable..

OK, now I think I will go for the WE..
Have a nice evening and sorry again for the misunderstanding..

Mirko.

mirkoflchtt:
Hi Nick, All,
Back to the subject about this proposed patch for auto-negotiation I enhanced in these days the patch in order to solve the main issues you mentioned: so thanks for pointing me out to the weaknesses of my initial patch !
Now I kept the protocol auto-negotiation feature has optional feature that could be enabled by only invoking specific new API. So the default behavior of the library is exactly the same as before: this should ensure proper unit tests are not broken any more. I merged-in also latest 2.6 version of the library from upstream and checked my derived version still work. 
If you are interested I could than propose a new pull request or if not I will close this one.. 
Latest changes I did are highlighted  here above if you want to have a look..

PR #124, opened 2016-01-30
Adding #ifndef blocks around commonly modified values

anonymouse64:
PubSubClient.h : adding #ifndef blocks around the commonly modified #defines to allow the values to be overridden by -D options when compiling. This allows the values to changed without modifying the library source files.


knolleary:
Hi, thanks for the contribution. There is already an identical pullrequest from a couple days ago waiting for me to deal with. So, going to close your one but not because we don't want to do this change, just that #119 got there first.

anonymouse64:
Ah I didn't even see that one...

Issue #156, opened 2016-04-24
potential buffer overflow and fix


edwin-oetelaar:
Extra check in readPacket.
Prevent malicious server from overflowing client buffer.
I think you should implement this.
Best regards,
Edwin van den Oetelaar

```

diff --git a/src/PubSubClient.cpp b/src/PubSubClient.cpp
index 9658c4a..b320dac 100755
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -244,6 +244,9 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
         buffer[len++] = digit;
         length += (digit & 127) * multiplier;
         multiplier *= 128;
+        /* check for protocol error */
+        if (multiplier > 0x200000)
+            return 0; /* invalid size, too large, more than 4 bytes in size */
     } while ((digit & 128) != 0);
     *lengthLength = len-1;
```

knolleary:
Hi, thanks for raising this. Yes, technically this is a check we should have. Will look to add it next time I'm making changes.

knolleary:
Fixed by 
https://github.com/knolleary/pubsubclient/commit/54be6e87db16d7dbc60d4c4674566ec08b0115c2
and
https://github.com/knolleary/pubsubclient/commit/4daba0ae5c11cca4da2fd98a1ba4fe0b490a4a86

PR #158, opened 2016-04-28
Fix Issue #156

Newman101:
Added prevention for buffer overflow in `src/PubSubClient.cpp`. Fixes issue #156.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -246,7 +246,9 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
         multiplier *= 128;
         /* Prevent buffer overflow */
         if (multiplier > 0x200000) {
-          return 0;
+          this->_state = MQTT_CONNECTION_LOST;
+          _client->flush();
+          _client->stop();
         }
     } while ((digit & 128) != 0);
     *lengthLength = len-1;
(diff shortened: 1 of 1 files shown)


knolleary:
Hi, thanks for this.

Simply returning 0 isn't sufficient. The client will then try to read the next packet... but we're no longer aligned to packet boundaries and there's no hope of recovering that.

If this condition is detected, the client needs to disconnect.

Newman101:
Sure. I'll make this disconnect then. Thanks for notification.

Newman101:
Do you wish the client to produce an error message upon disconnection?

edwin-oetelaar:
I personally do not find this very elegant or logical in program flow.
You close the TCP connection inside a reading loop and do not bail out cleanly even when you know of the error condition.
You depend on the readByte() function to return false on the closed connection and then make it return 0..
On detecting the error : Closing the connection, setting the state to 'protocol error' and directly returning 0 would be much cleaner.
I would even prefer the readPacket() function to return a signed int instead of an unsigned one, that way a negative number could indicate all kinds of errors... like connection-lost, protocol-error, whatever.
The readPacket() function is only used in 2 places, so the impact would be low.

Greetings,
Edwin van den Oetelaar

knolleary:
@edwin-oetelaar Agreed - the disconnection needs to done via a more graceful flow.

@Newman101 fyi, I don't get notified when you add a commit to the pull request, which is why I hadn't spotted your changes. Please add a comment if you have changed the pr.

Newman101:
Sure, thanks for the notification. I can amend the current PR to correspond more graceful flow.  At the moment the error detection does not return ```0```.

knolleary:
I pushed an alternative fix for this back in July - and have just refined it (it was still off-by-one) with https://github.com/knolleary/pubsubclient/commit/4daba0ae5c11cca4da2fd98a1ba4fe0b490a4a86

Issue #110, opened 2016-01-12
Function to change the MQTT packet size outside of the library


marcoschwartz:
Hello, great library! For several of my projects I'd need to change the MQTT packet size, but I need to do it from outside of your library.

Indeed, now I'd need to change MQTT_MAX_PACKET_SIZE inside your library, which is not very convenient. A function or a way to do that without changing your library would be great! Thanks.

skorokithakis:
+1, I'm using PlatformIO which keeps a common cache for all libraries and this is presenting a big problem for me. An easier way to configure this would be great.

lunanigra:
I need this too... 128 seems far too low.

We are using DeviceHub. Based on the long API key and Devic ID length you easily exceed 128.
So, perhaps set MQTT_MAX_PACKET_SIZE to 256 or better make it configurable.

Thanks :-)

knolleary:
So the reason it's fixed is because we preallocate the memory at compile time. Given the limited memory of the original devices this targeted, I felt it was import to do this so potential memory issues could be caught then, and not at runtime (which are so much harder to debug). I also didn't want to set a default value that meant a lot of memory was pegged but went unused if you were only sending/receiving simple messages.

Any default value will have pros and cons.

Perhaps the answer is to move to fully dynamic memory allocation, but isn't a straight forward choice.

skorokithakis:
Could it be made optional with an ifdef so it can be defined at the preprocessor level perhaps, at least?

knolleary:
I never found a way to allow an arduino sketch to ifdef its own value to override the library - all to do with how the  Arduino tool chain preprocesses the app's files.

skorokithakis:
I use PlatformIO, which can do that (or, if not, I'm sure @ivankravets will approve a PR from me for it).

ivankravets:
@knolleary what is problem with `#ifdef` https://github.com/knolleary/pubsubclient/blob/master/src/PubSubClient.h#L23?

```c
#ifndef MQTT_MAX_PACKET_SIZE
#define MQTT_MAX_PACKET_SIZE 128
#endif
```

[PlatformIO](http://platformio.org) users have full control of build process and can pass own [build_flags](http://docs.platformio.org/en/latest/projectconf.html#build-flags). For example, `platformio.ini`

```ini
[env:myboard]
platform = espressif
framework = arduino
board = nodemcu
build_flags = -DMQTT_MAX_PACKET_SIZE=256
```

[...]

PR #165, opened 2016-05-20
Allow external buffer allocation

vicatcu:
Addresses https://github.com/knolleary/pubsubclient/issues/110

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -606,5 +606,6 @@ int PubSubClient::state() {
 PubSubClient& PubSubClient::setPacketBuffer(uint8_t * pBuf, uint16_t length){
   MQTT_MAX_PACKET_SIZE = length;
   buffer = pBuf;  
+  return *this;
 }
 
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -11,6 +11,7 @@ PubSubClient::PubSubClient() {
     this->_state = MQTT_DISCONNECTED;
     this->_client = NULL;
     this->stream = NULL;
+    setPacketBuffer(&(this->_buffer[0]));
     setCallback(NULL);
 }
 
@@ -18,6 +19,7 @@ PubSubClient::PubSubClient(Client& client) {
     this->_state = MQTT_DISCONNECTED;
     setClient(client);
     this->stream = NULL;
+    setPacketBuffer(&(this->_buffer[0]));    
 }
 
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client) {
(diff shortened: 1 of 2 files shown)


jonnor:
Tests are failing, `src/receive_spec.cpp:108:18: error: ‘MQTT_MAX_PACKET_SIZE’ was not declared in this scope`.
The meaning of `MQTT_MAX_PACKET_SIZE` should be preserved, as it is quite likely used by external code.

src/PubSubClient.cpp
```diff
@@ -11,94 +11,108 @@ PubSubClient::PubSubClient() {
     this->_state = MQTT_DISCONNECTED;
     this->_client = NULL;
     this->stream = NULL;
+    setPacketBuffer(&(this->_buffer[0]));
```
jonnor: Pretty sure this can be just `setPacketBuffer(this->_buffer)`.

src/PubSubClient.cpp
```diff
@@ -588,3 +602,10 @@ PubSubClient& PubSubClient::setStream(Stream& stream){
 int PubSubClient::state() {
     return this->_state;
 }
+
+PubSubClient& PubSubClient::setPacketBuffer(uint8_t * pBuf, uint16_t length){
+  MQTT_MAX_PACKET_SIZE = length;
```
jonnor: Looks like assignment to a global/define due to the uppercasing. The instance variable should probably follow the regular conventions (camelCase). For instance `bufferLength`

src/PubSubClient.h
```diff
@@ -83,7 +83,9 @@
 class PubSubClient {
 private:
    Client* _client;
-   uint8_t buffer[MQTT_MAX_PACKET_SIZE];
+   uint8_t * buffer;
+   uint16_t MQTT_MAX_PACKET_SIZE;
+   uint8_t _buffer[_MQTT_MAX_PACKET_SIZE];
```
jonnor: Minor, but `_buffer` is scarily close to `buffer`, makes it easy to accidentially misuse. Think this should have a more verbose/ugly name. Like `_staticInternalBuffer`.

src/PubSubClient.h
```diff
@@ -121,6 +123,7 @@ class PubSubClient {
    PubSubClient& setCallback(MQTT_CALLBACK_SIGNATURE);
    PubSubClient& setClient(Client& client);
    PubSubClient& setStream(Stream& stream);
+   PubSubClient& setPacketBuffer(uint8_t * pBuf, uint16_t length = _MQTT_MAX_PACKET_SIZE);
```
jonnor: Having a default value here makes it easy for people to forget passing the length. Think it should be dropped, even if it means a bit of code duplication in the default constructors.

PR #169, opened 2016-06-14
Use random MQTT client ID in example sketch

igrr:
Many users load the sketch having only changed WiFi SSID and password. When multiple users attempt to connect to the same broker using same client ID, they get rejected. Currently the chances of connecting to broker.mqtt-dashboard.com using "ESP8266Client" ID are fairly slim. This change adds a random number to the client ID, increasing chances of connection for new users of this library.


jonnimoura:
I had the same error and only solved when I read your comment.

Thanks!!

knolleary:
Good idea... will add to the other example sketches as well

PR #170, opened 2016-06-22
Rework packet handling

mknjc:
I've started to add a new callback which gets the payload length and the topic before the data appears in the stream.
But first I cleaned up the packet handling up. The new callback will be in a separate pull request.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -125,7 +125,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
         if (result == 1) {
             nextMsgId = 1;
             // Leave room in the buffer for header and variable length field
-            uint16_t length = 5;
+            uint32_t length = 5;
             unsigned int j;
 
 #if MQTT_VERSION == MQTT_VERSION_3_1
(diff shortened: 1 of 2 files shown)


knolleary:
Hi,

please can you summarise what you're changing and what benefits it brings?

mknjc:
Hi,

quick c&p from my commit:
split the code out of readPacket and loop into multiple functions
don't use the buffer for headers
change length variables to uint32_t, mqtt packets can be 2^28 = 268435456 bytes long
some more checks for invalid protocol
don't copy the topic around
throw away the code for MQTTPINGREQ packets, only the clients sends them

the major part is the cleanup of the loop function, most of the packet handling moved into the handlePublishPacket function.
because the topic handling and the payload handling aren't interleaved anymore it's now easy to add a callback before reading the payload.

knolleary:
Right - but why is that useful? What scenario does having a callback before the payload has been read enable?

mknjc:
If I wan't to save a payload bigger then my free ram in my flash I don't to write every message into the flash and check after the stream writing in the callback if my message is the expected. Even if I subscribe the topic I'm interested in immediate after adding the stream the server might have a publish in the queue  which get served first. The only sane way is getting the topic and size before reading the payload.

knolleary:
That is the purpose of the MAX_PACKET_SIZE setting - why do you need a second mechanism for protecting against payloads that are too big?

mknjc:
I wan't to receive the payload but can't handle it in my ram, so I have to use the stream interface, but currently the stream receives the data before the callback get called.

PR #173, opened 2016-06-26
Reduce stack usage in loop(). no data duplication on stack

edwin-oetelaar:
remove copy of topic string onto stack before giving it to callback()
Just move it one byte in buffer to add space for 'C' string end \0x00 same result, no stack needed.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -306,12 +306,10 @@ boolean PubSubClient::loop() {
                 uint8_t type = buffer[0]&0xF0;
                 if (type == MQTTPUBLISH) {
                     if (callback) {
-                        uint16_t tl = (buffer[llen+1]<<8)+buffer[llen+2];
-                        char topic[tl+1];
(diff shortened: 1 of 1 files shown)


Suxsem:
Umh -1, payload isn't always a string

Suxsem:
Moreover, see mqttDataCb() comments in https://github.com/Suxsem/pubsubclient/blob/master/examples/mqtt_esp8266/mqtt_esp8266.ino

edwin-oetelaar:
Topic can not contain embedded '0x00' chars.
4.7.3 Topic semantic and usage
http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718109
The fix reduces the stack usage, it no longer copies the full topic onto the stack.
I am not touching any other payload.
Please read the code before commenting.

Suxsem:
You're right I read payload instead of topic sorry. 
I will review the proposal more carefully and report back here

knolleary:
Looks sensible to me

Issue #175, opened 2016-06-30
Make MQTT_KEEPALIVE modifiable at runtime


richjoyce:
It's well known that the Arduino IDE doesn't support changing defines at compile time in libraries. Issue #110 addresses this for another define, but I think the cons brought up in that conversation don't apply to the `MQTT_KEEPALIVE` define. It is only used in `connect` so I believe this would be simple to add as an optional argument to `connect`, and wouldn't take up any additional memory space.

I'm happy to submit a PR if this would theoretically get accepted. 

This argument also mostly applies to `MQTT_SOCKET_TIMEOUT`, but it would need to be stored in the class for `readByte` to use it. I could make this change at the same time.

knolleary:
Agreed - making this dynamic is a significantly smaller overhead than the buffer size.

Happy to look at a PR for it

knolleary:
Have just pushed a fix that adds `setKeepAlive`. Thanks for proposing this change - sorry it took so long to come to a resolution.

PR #177, opened 2016-07-05
Allow custom values for MQTT_KEEPALIVE, MQTT_SOCKET_TIMEOUT

karasinski:
Addresses issue https://github.com/knolleary/pubsubclient/issues/175. I'm not sure how you feel about optional values in the header declaration, so if this is annoying for you feel free to just take the first commit.


knolleary:
Hi, the problem here is rather than just make them configurable, you've removed the constants from the header. That is a breaking change so I can't merge it. There's no reason not to Jeep those constants as the place the default values are defined if a sketch doesn't provide its own. That would be entirely backward compatible.

knolleary:
I've just pushed a change that adds `setKeepAlive()` - allowing the constant to be overridden by the user's sketch. I'll be making a similar change for the socket timeout.

Given all of the other changes that have happened since this PR was first raised, and my comments above, I'm going to close this.

Thank you for taking the take to propose the change - my apologies it has taken so long to come to any resolution.

PR #178, opened 2016-07-05
Update documentation for new keepAlive and socketTimeout

richjoyce:
Documentation for PR #177, resolving issue #175


knolleary:
See my comment on #177 - removing the header constants is not acceptable

knolleary:
No longer required as the other PRs have been closed. Will be making separate updates to the dos for the new functions once the new version is published.

PR #194, opened 2016-08-24
Add homepage URL; update supported development platforms

ivankravets:


knolleary:
Hi - why change the platforms to *?

ivankravets:
Hi @knolleary,

I see that you specified `architectures=*` in `library.properties`. With the which development platforms ([see list](https://github.com/platformio?utf8=✓&query=platform-)) this library is compatible?

P.S: PlatformIO supports `library.properties` manifest. If you don't want to support `library.json`, we can remove it. I see that your library doesn't have dependencies and don't require custom build options that [library.json](http://docs.platformio.org/en/latest/librarymanager/config.html) provides. The only one difference is that `keywords` will not be reflected in PlatformIO Registry.

PR #206, opened 2016-10-04
Add QOS 1 Publish capability 

dagnall53:
I have added / hacked in a simple QOS 1 publish capability.
It adds a new boolean "publish_Q1" that includes a "save the last message" in variables: _SentQOS1buffer[MQTT_MAX_PACKET_SIZE],  _SentQOS1Topic[40]; and  _SENTQOS1Length;    

The topic length at 40 was ok for me, but should probably be increased. 
I had trouble at first with saving to these "last message/topic" variables so that i could resend using the same "Publish_Q1" code, 
A better approach would probably be to make a single stored message with the header, msg id and payload and then explicitly resend this if the PUBACK is not detected.? I was not sure how to achieve this simply.

It is limited to a single "stored message".

Cheers
Dagnall

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -12,14 +12,14 @@ PubSubClient	KEYWORD1
 # Methods and Functions (KEYWORD2)
 #######################################
 
-connect 	KEYWORD2
+connect 	  KEYWORD2
 disconnect 	KEYWORD2
-publish 	KEYWORD2
+publish 	  KEYWORD2
 publish_P 	KEYWORD2
-publish_Q1 KEYWORD2
+publish_Q1  KEYWORD2
 subscribe 	KEYWORD2
 unsubscribe 	KEYWORD2
-loop 	KEYWORD2
+loop 	      KEYWORD2
 connected 	KEYWORD2
 setServer	KEYWORD2
 setCallback	KEYWORD2
--- a/keywords.txt
+++ b/keywords.txt
@@ -16,6 +16,7 @@ connect 	KEYWORD2
 disconnect 	KEYWORD2
 publish 	KEYWORD2
 publish_P 	KEYWORD2
+publish_Q1 KEYWORD2
 subscribe 	KEYWORD2
 unsubscribe 	KEYWORD2
 loop 	KEYWORD2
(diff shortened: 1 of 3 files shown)


knolleary:
Hi @dagnall53 

there has been lots of discussion around adding qos>0 to this library. Not sure if/when I'll get the chance to do it properly. 

This library tries to be very conscious of its memory foot print. By adding a second internal buffer, this pull request instantly doubles the space required. This client is also explicitly a clean-session only client (which is a separate thread of discussion). It is inappropriate for this client to be automatically reattempting publishes after a reconnect.

If an application sends a second qos1 message before the first has been acknowledged it isn't clear to me your logic works.

dagnall53:
Thanks,

1 Memory.. I understand and appreciate the coding for minimum memory, but the large last payload sent store is not "too huge" IMHO in comparison to the memory on the processor I'm using (ESP12E).. And an  # IF DEF at the beginning might allow the Q0S1 features to be switched OFF if minimal size is essential ?

2 Not sure I understand the Clean session issue, but presumably you mean if we lose connection, we would lose synch and PUBACKs / Publish's could get horribly lost.. ?....  Yes I can see that would happen. I will try to read up about that other thread.

3, I agree re multiple qos1 messages. My code explicitly does not work if two qos1 messages are sent before the first gets PUBACK. .. But experimentally I have been getting PUBACK back in less than 100ms, (5-35ms max seen so far)  and the send rate of my application is relatively low.

But, am I correct that MQQT requires all QOS1 communications to be in strict msgID order? If so, a stack of "pending qos1 sends"  might be an appropriate approach to accommodate multiple qos1 "in flight".

dagnall53:
Hold on any merging, I have just discovered that my code was broken when I removed the debug.. I will find out what I did and correct it.

knolleary:
@dagnall53 no problem - as I've said, I have concerns over the general approach here, so I won't be rushing to merge it until I get some more thought into it.

dagnall53:
Thanks, I have just been doing experiments, and Is the "clean-session only client " issue one associated with losing connection and then reconnecting.. If so then I just saw it and it certainly screwed up my comms and I started sending lots of re-sends..
also ..Is there a "Duplicate" bit I should be setting somewhere in my re-sends?

knolleary:
The duplicate flag was removed in 3.1.1.

f0nt4:
Would be nice to see this one merged. If memory really is a concern you might want to split it into 2 different packages?

dianjixz:
Repackaging high-quality messages into a package that takes care of both memory and functionality.Why not

dagnall53:
I'm not working with my branch anymore, as I prefer to use the standard code, and found that in my application I did not really need better QOS in practice. 
But feel free to make  / update / your own branches if you want. .

PR #218, opened 2016-12-11
Support for passing arguments to a callback

adaszek:
Hi,

I've added a possibility to pass an argument into callback function. It may be useful if someone wants to organize subscription handlers into more object-oriented way without unnecessary usage of global variables (e.g. handlers of topic changes encapsulated as separate class).

Usage is similar to arg parameter from pthread_create().

It doesn't break legacy functionality.

### DIFF
--- a/src/PubSubClient.h
+++ b/src/PubSubClient.h
@@ -76,8 +76,10 @@
 #ifdef ESP8266
 #include <functional>
 #define MQTT_CALLBACK_SIGNATURE std::function<void(char*, uint8_t*, unsigned int)> callback
+#define MQTT_CALLBACK_SIGNATURE_WITH_ARG std::function<void(char*, uint8_t*, unsigned int, void*)> callbackWithArg
 #else
 #define MQTT_CALLBACK_SIGNATURE void (*callback)(char*, uint8_t*, unsigned int)
+#define MQTT_CALLBACK_SIGNATURE_WITH_ARG void (*callbackWithArg)(char*, uint8_t*, unsigned int, void*)
 #endif
 
 class PubSubClient {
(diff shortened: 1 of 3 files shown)


knolleary:
I am not familiar with pthread_create.

Could you add a concrete example of how this would be used?

adaszek:
Small example how it can be utilized is available in the latest commits.

Suxsem:
The idea is good but a new example sketch that simply shows the new callback in action (without all the test stuff overhead) should be provided, in my opinion

adaszek:
@Suxsem I've just added and example you asked for. It was tested only on my homebrew arduino with NRF24L01+ but it should not matter. Unfortunately, I don't have ethernet shield to make this example simpler.

adaszek:
@knolleary Hi, any chances to have it merged?

src/PubSubClient.h
```diff
@@ -88,7 +90,9 @@ class PubSubClient {
    unsigned long lastOutActivity;
    unsigned long lastInActivity;
    bool pingOutstanding;
-   MQTT_CALLBACK_SIGNATURE;
+   MQTT_CALLBACK_SIGNATURE;  
+   MQTT_CALLBACK_SIGNATURE_WITH_ARG; 
+   void* arg;
```
knolleary: This should have a more descriptive name - `arg` is too generic to know what its related to.

examples/mqtt_rf24ethernet_and_callback_with_argument/mqtt_rf24ethernet_and_callback_with_argument.ino
```diff
@@ -0,0 +1,193 @@
+/*
```
knolleary: This feels like quite a complex example. Saying 'its up to the developer to decide about the design of topic handlers' doesn't give the develop much help to understand how to use this feature.

The example also needs more documentation to explain what its actually doing - it's taken me a while to piece together how it works.

jaecktec:
here is a way how to build it object orientated (thats why I wanted the functional way on the esp32

jaecktec:
here is a way how to build object orientated: 
https://github.com/jaecktec/mqtt-device-registry-client/blob/master/DeviceRegistryDevice.cpp
Thats why I wanted the functional way for the esp32

jaecktec:
here is a way how to build object orientated:
https://github.com/jaecktec/mqtt-device-registry-client/blob/master/DeviceRegistry.cpp#L75
Thats why I wanted the functional way for the esp3

PR #251, opened 2017-03-03
Added SSL support.

diegopx:
A new constructor (available only on the ESP8266) allows the user to verify the SSL certificate fingerprint. Also, to avoid read/write collisions in axTLS buffer, mqtt.loop processes all incoming messages instead of just one.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -313,62 +313,67 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
 
 boolean PubSubClient::loop() {
     if (connected()) {
-        unsigned long t = millis();
-        if ((t - lastInActivity > MQTT_KEEPALIVE*1000UL) || (t - lastOutActivity > MQTT_KEEPALIVE*1000UL)) {
-            if (pingOutstanding) {
-                this->_state = MQTT_CONNECTION_TIMEOUT;
-                _client->stop();
-                return false;
-            } else {
-                buffer[0] = MQTTPINGREQ;
-                buffer[1] = 0;
-                _client->write(buffer,2);
-                lastOutActivity = t;
-                lastInActivity = t;
-                pingOutstanding = true;
-            }
-        }
-        if (_client->available()) {
-            uint8_t llen;
-            uint16_t len = readPacket(&llen);
-            uint16_t msgId = 0;
-            uint8_t *payload;
-            if (len > 0) {
-                lastInActivity = t;
-                uint8_t type = buffer[0]&0xF0;
-                if (type == MQTTPUBLISH) {
-                    if (callback) {
-                        uint16_t tl = (buffer[llen+1]<<8)+buffer[llen+2]; /* topic length in bytes */
-                        memmove(buffer+llen+2,buffer+llen+3,tl); /* move topic inside buffer 1 byte to front */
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for this. Don't know how soon I'll be able to look at it properly to get it merged. Have you got a link to the relevant ESP8266 docs for the TLS support?

One observation; the client allows publishing from the message-arrived callback. According to the comments you've added, doing so would trash the axTLS buffer and you'd lose messages?

diegopx:
In their master as it is now (esp8266/Arduino@04df3adb54a73f13960738fa10a0a0b680f68cf0), yes, trying to publish inside the message handler would trash the buffer and any further read would return garbage. Then, what usually happens is PubSubClient halts (or times out) because the read message length is a arbitrary large number and the socket will wait to get that many bytes.

I've fix that in my fork of Arduino, so WiFiClientSecure::write() returns an error instead of trashing the buffer (see pull request esp8266/Arduino#3019). Hence, mqtt.publish() might fail inside the message callback.
If there were no further messages on the SSL socket when reading the PUBLISH message packet, then the publish() would succeed as usual.

Since WiFiClientSecure uses axTLS, a general never-fail publish-in-onmessage is not possible without an additional buffer, because two messages could come in the same SSL packet, and the first one will never be able to publish without killing the buffer for the second.

diegopx:
As a sidenote, one disadvantage of the code is that since it loops reading messages, there is a possible case where the client is overflown with new messages and mqtt.loop() never returns, so the rest of the user code never executes. Although this situation is very extreme, it may still happen.

Only axTLS is half duplex, the internal ClientContext has its own (non-decrypted) buffer. So, the collisions only start when doing ssl_read(), which reads from the ClientContext into its own buffer and decrypts (i.e. the collisions do not start immediately when the data arrives to the device). Therefore, there are moments where there may still be a non-decrypted message in the ClientContext but the axTLS buffer is empty. These moments are ideal to exit from the loop.
In particular, calling WiFiClientSecure.available() gives the exact size remaining in the buffer, but induces a new ssl_read() if the buffer was empty. So I'll try and see if I can reduce the number of available() calls and caching such value so that the processing only consumes the "original buffer" (i.e. the buffer as of the moment mqtt.loop() was called) and does not keep on decrypting further messages. This way the rest of the user loop has a chance to run.

odelot:
hi @diegopx. btw did you get this behaivor while using WifiClientSecure with the mqtt server?

The WifiClientSecure.write method returns with success even though there is no internet connection? (so the device is connect to a router, sending data to a clouded mqtt server and even with no internet it returns success)

This behaivor leads to a memory problem and freezes the device after a while (the WifiClientSecure continues to allocate memory to encrypt the message even it does not send it).

lizaoreo:
Has this not been merged?

dbudwin:
@knolleary Any progress on this?  I really like this library and would like to be able to continue using it to securely subscribe to MQTT topics.

Enrico204:
@diegopx I noticed that you're using `verify` before checking for `result` variable to be equal to 1. I think that this might lead to crash when `connect` fails (I just tested it, and it seems to crash when MQTT server is not responding).

diegopx:
It's been a while, but from what I can see, you're correct; the connect return value should be checked before verifying.

PR #252, opened 2017-03-07
Update to support binary input for the user, password and payload

flemlion:
We intend to enable the use of encryption during the connection and for the payload.
MQTT as a binary protocol is able to support this, but the library not yet.
This update fills that gap.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -102,18 +102,26 @@ PubSubClient::PubSubClient(const char* domain, uint16_t port, MQTT_CALLBACK_SIGN
 }
 
 boolean PubSubClient::connect(const char *id) {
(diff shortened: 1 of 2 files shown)


NicksonYap:
@flemlion 
I was having trouble publishing raw binary data (byte by byte). 
The payload seems to end the moment '0' (null character) is detected.
So I looked for a PR/issue and found this.

However, your PR doesn't seem to have implemented sending binary data for publish.
I only see it for connect.

Please correct me if I'm wrong.

knolleary:
The client already supports publishing binary data - http://pubsubclient.knolleary.net/api.html#publish3

NicksonYap:
@knolleary my mistake. i was using an older version of the library.
Now it;s flawless!
Thanks a lot!

PR #261, opened 2017-03-27
Add the esp32 example

0x1abin:


knolleary:
HI, thanks for this.

can you describe what is specific to this example regarding the esp32? Looking at the code I don't immediately see anything that the other examples don't cover already.

bbx10:
Here are the trivial changes make mqtt_esp8266.ino work with the latest arduino-esp32. 1) Every include file with "ESP8266" in the name has been renamed to remove the "ESP8266". 2) gcc does not like using %ld with an int. Either value must be a long => %ld or int => %d. I tested against mosquitto running on raspbian.

```
Linux raspi2 4.9.24-v7+ #993 SMP Wed Apr 26 18:01:23 BST 2017 armv7l GNU/Linux
Package: mosquitto
Version: 1.3.4-2
Architecture: armhf
```

```C
--- ~/Sync/pubsubclient/examples/mqtt_esp8266/mqtt_esp8266.ino	2017-05-20 07:54:46.512756332 -1000
+++ ./mqtt_esp8266.ino	2017-05-20 08:50:23.588450342 -1000
@@ -1,8 +1,8 @@
 /*
- Basic ESP8266 MQTT example
+ Basic ESP8266/ESP32 MQTT example
 
  This sketch demonstrates the capabilities of the pubsub library in combination
- with the ESP8266 board/library.
+ with the ESP8266/ESP32 board/library.
 
  It connects to an MQTT server then:
   - publishes "hello world" to the topic "outTopic" every two seconds
@@ -23,7 +23,11 @@
 
 */
 
+#ifdef ESP8266
 #include <ESP8266WiFi.h>
+#else
+#include <WiFi.h>
+#endif
 #include <PubSubClient.h>
 
@@ -124,7 +128,7 @@
   if (now - lastMsg > 2000) {
     lastMsg = now;
     ++value;
-    snprintf (msg, 75, "hello world #%ld", value);
+    snprintf (msg, sizeof(msg), "hello world #%d", value);
     Serial.print("Publish message: ");
     Serial.println(msg);
     client.publish("outTopic", msg);
```

examples/mqtt_esp32/mqtt_esp32.ino
```diff
@@ -0,0 +1,133 @@
...
+  for (int i = 0; i < length; i++) {
+    Serial.print((char)payload[i]);
+  }
+  Serial.println();
+
+  // Switch on the LED if an 1 was received as first character
+  if ((char)payload[0] == '1') {
+    digitalWrite(BUILTIN_LED, LOW);   // Turn the LED on (Note that LOW is the voltage level
+    // but actually the LED is on; this is because
+    // it is acive low on the ESP-01)
```
tobymurray: Trivial, but spelling mistake (`acive` -> `active`). Also, noting that I just tried an ESP32 for the first time on a ESP-WROOM-32, it looks like it's active high? Not sure what the ESP-01 is, perhaps different.

PR #270, opened 2017-04-12
Make Python files compliant with PEP8, except for E501

pacm93:
Hi, I am using this client for the ESP8266 and when I ran flake8 for my Python files, it also returned several warnings regarding the test scripts of this library.
Besides cosmetic changes, I also made the prints compatible with Python3, removed 'import serial' because it was not being used and modified the logic for verifying the 'ino' library.

### DIFF
--- a/tests/testcases/mqtt_basic.py
+++ b/tests/testcases/mqtt_basic.py
@@ -1,43 +1,39 @@
 import unittest
 import settings
-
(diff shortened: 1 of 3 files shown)


knolleary:
Thanks!

pacm93:
You are welcome!

PR #275, opened 2017-04-20
Added reference to PubSubClient object to callback prototype

chessweb01:
As it is, the callback function relies on an external/global instance of
the PubSubClient object if it wants to use the object for instance for
publishing.

It is preferable to be able to pass a reference to the PubSubClient
instance into the callback function when it is called.

To that end the callback signature and the calls to the callback in the
loop method had to be changed.


chessweb01:
Forgot to adapt the tests to the new callback protoype.

knolleary:
Hi,

This change will break all existing sketches, so i can't merge it as is.

The is also pr #218 that's in a similar area but has been done in a way not to break things. I think that pr would satisfy your requirement?

chessweb01:
You are right, #218 achieves the same in a more general fashion. So I'm going to stick to my local modification until you pull #218.

PR #282, opened 2017-04-30
Dynamic packet size

mrdunk:
Allow dynamic allocation of buffer size so rebuilding library is not required if user needs greater than MQTT_MAX_PACKET_SIZE (128) byte packet size.
See discussion at https://github.com/knolleary/pubsubclient/issues/110

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -12,12 +12,16 @@ PubSubClient::PubSubClient() {
     this->_client = NULL;
     this->stream = NULL;
     setCallback(NULL);
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 
 PubSubClient::PubSubClient(Client& client) {
     this->_state = MQTT_DISCONNECTED;
     setClient(client);
     this->stream = NULL;
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client) {
@@ -25,12 +29,16 @@ PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client) {
     setServer(addr, port);
     setClient(client);
     this->stream = NULL;
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, Client& client, Stream& stream) {
     this->_state = MQTT_DISCONNECTED;
     setServer(addr,port);
     setClient(client);
     setStream(stream);
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, MQTT_CALLBACK_SIGNATURE, Client& client) {
     this->_state = MQTT_DISCONNECTED;
@@ -38,6 +46,8 @@ PubSubClient::PubSubClient(IPAddress addr, uint16_t port, MQTT_CALLBACK_SIGNATUR
     setCallback(callback);
     setClient(client);
     this->stream = NULL;
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 PubSubClient::PubSubClient(IPAddress addr, uint16_t port, MQTT_CALLBACK_SIGNATURE, Client& client, Stream& stream) {
     this->_state = MQTT_DISCONNECTED;
@@ -45,6 +55,8 @@ PubSubClient::PubSubClient(IPAddress addr, uint16_t port, MQTT_CALLBACK_SIGNATUR
     setCallback(callback);
     setClient(client);
     setStream(stream);
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
 }
 
 PubSubClient::PubSubClient(uint8_t *ip, uint16_t port, Client& client) {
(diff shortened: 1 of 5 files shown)


knolleary:
Hi, thanks for this. It is, on face value, a reasonable change. It's just needs me to have sufficient spare time to review in detail and go through the process of merging and releasing etc. Can't give you an outlook for that right now.

mrdunk:
All understood.
Let me know if i can be of any assistance.

thanks,
dunk.

hessijames79:
+1, also interested in dynamic packet size

jonnor:
Any news on this? Right now when using this library have to tell everyone to download a forked copy or hack the existing define in their library folder...

knolleary:
No news - will try to get to it in the next week.

Aat0304:
+1, more than interested in dynamic packet size. Thx

MarkoHoe:
+1

marcostrullato:
Interested as well.

Thanks

gilles906:
+1 Interested as well.

Thanks

svanscho:
yep +1 here, seems like some of my messages get lost if they are bigger than 128 bytes

coogle:
+1 Please merge this! It's a completely unnecessary burden to ask people to adjust a compile-time setting to build a library when it's got a dependency on this library. Looking at the code it seems completely backward compatible.

mew1033:
Definitely interested in this. Thanks for the PR!

gerhardgoosen:
do i have to pull this code manually into my cloned repo.. not seeing this in the master branch

src/PubSubClient.cpp
```diff
@@ -12,93 +12,125 @@ PubSubClient::PubSubClient() {
     this->_client = NULL;
     this->stream = NULL;
     setCallback(NULL);
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
```
Titantompa: Woudn't it be better to move the buffer allocation to a support function, rather than duplicate the same code in all constructors?

src/PubSubClient.cpp
```diff
@@ -12,93 +12,125 @@ PubSubClient::PubSubClient() {
...
     setServer(domain,port);
     setCallback(callback);
     setClient(client);
     setStream(stream);
+    this->buffer_size = MQTT_MAX_PACKET_SIZE;
+    this->buffer = (uint8_t*)malloc(MQTT_MAX_PACKET_SIZE);
+}
+
+PubSubClient::~PubSubClient() {
+  free(this->buffer);
```
Titantompa: This should check if this->buffer is NULL or not since setBufferSize() can set it to NULL.

src/PubSubClient.cpp
```diff
@@ -586,3 +618,13 @@ PubSubClient& PubSubClient::setStream(Stream& stream){
 int PubSubClient::state() {
     return this->_state;
 }
+
+boolean PubSubClient::setBufferSize(uint16_t size) {
+  this->buffer = (uint8_t*)realloc(this->buffer, size);
```
Titantompa: This should use a temp variable. If realloc fails it will return a NULL pointer without freeing the original memory. Thus, if NULL is returned this method should not make any change to this->buffer since it is still valid.

tests/src/receive_spec.cpp
```diff
@@ -242,6 +279,7 @@ int main()
     test_receive_stream();
     test_receive_max_sized_message();
     test_receive_oversized_message();
+    test_resize_buffer();
```
Titantompa: Since the merge request contains functionality to dynamically change the buffer size it would be suitable to have tests that verify that this works well. E g:

- What happens if setBufferSize() fails to allocate the memory?
- If the buffer size is shrunk will it fail (appropriately) to receive a message that was within the prior limits?
- And the reverse, if it fails to receive a message due to size will increasing the buffer size accordingly help?

Titantompa (changes requested):
This was done in a hurry so there may be other comments.

src/PubSubClient.cpp
mrdunk: i agree in principal but
i was trying to match the existing code style. (ie, explicitly calling everything in each constructor.)

@knolleary any opinion?

src/PubSubClient.cpp
mrdunk: Nope. according to the spec std::free() should treat a null pointer as a no-op.
from http://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf :
> The free function causes the space pointed to by ptr to be deallocated, that is, made available for further allocation. If ptr is a null pointer, no action occurs.

(it makes sense because std::malloc returns null pointer on failure.)

src/PubSubClient.cpp
mrdunk: yes. this makes sense.

@knolleary is this pull request likely to ever be merged? is it worth me doing this?

tests/src/receive_spec.cpp
mrdunk: yes, all make sense. Let's see if this is likely to be merged first.

src/PubSubClient.cpp
Titantompa: Ah, sweet, I haven't really coded C/C++ in ages and back then it was a no-no since freeing a null pointer would cause some environments to go belly up. :-)

coogle:
I'm curious if this is going to be merged? My project (http://github.com/ThisSmartHouse/CoogleIOT) relies on PubSubClient for it's MQTT and the lengthy topics the framework uses causes problems (specifically with my heartbeats of devices). Obviously we can side-step this issue with a compile-time flag but what can we do to actually get this dynamic allocation code into the library proper?

sskorol:
heavy plus for this feature (as well as for other similar configuration stuff)!..

forcing people to direct sources' modification means additional maintenance effort delegation to client code... so instead of just clicking "install" in a library manager to retrieve a new version, people have to keep an up to date overridden sources in their own libraries or frameworks... that might be quite disappointing... don't you think so?..

but anyway thanks for the great lib!..

TheCoolBrit:
Any news on this?
I am not very up to date on programming so using Arduino IDE  with a TTGO T-Display to display 240x135 pictures via MQTT using PubSubClient
Getting even this far has been a long winding slog for me so Please
Can anyone show me how to change the PubSubClient to allow #define  MQTT_MAX_PACKET_SIZE 64800
Thank you in anticipation.

coogle:
At this point I don't understand why it hasn't been merged, other than the owner of the project just doesn't want to do it 👎

knolleary:
I agree this is a long overdue feature.

The challenge has always been finding the time and energy to do the work needed to accept a change like this - especially as it's a feature I have little personal need for. I've also been fairly preoccupied with one of my other OSS projects...

That said, I have recently started dusting off my hardware so I can test some of the proposed enhancements to the library. This one is high on the list... in fact, I may well have a play with it this evening...

TheCoolBrit:
What do you make to this possible alternative >
https://github.com/marvinroger/async-mqtt-client

Appreciate any comments

knolleary:
Well, it took me 3 years, but we got there eventually. Thanks for your patience @mrdunk.

Aat0304:
Cool,  Thanks, AatVerzonden vanaf mijn mobiele telefoon-------- Oorspronkelijk bericht --------Onderwerp: Re: [knolleary/pubsubclient] Dynamic packet size (#282)Van: knolleary Aan: knolleary/pubsubclient Cc: Aat0304 ,Comment 
Merged #282 into master.

mrdunk:
oh cool, just saw this made it in.
thanks knolleary!
i know it's hard to spend time maintaining stuff you don't have much use for.

PR #305, opened 2017-07-21
Fix: Rubbish data in received payload, due to previous publishing

hongquan:
Issue: My Arduino both do pushlish and subscribe to different topic.
After publishing, there appears some rubbish data in consecutive  received payload.


knolleary:
Can you provide an example of what you are seeing? There really is no need to empty the buffer like this.

hongquan:
We checked again.

It turns out that the "rubbish" data is because we subscribed to more topics (via '#') than we are interested in.

Issue #308, opened 2017-07-24
First run mqttclient.loop() takes 15+ seconds


gjt211:
I am using this library with a WiFi secure connection to my own dedicated server running Mosquitto.

After Wi-Fi is connected and secure connection made to my server, In my main loop, I test if mqtt is connected.
`if (!mqttclient.connected()) {
    reconnect();
  }`
Then in my reconnect function (which is really both a connect & reconnect function), I have the following code
```
while (!mqttclient.connected()) {
Serial.print("State:");
Serial.println(mqttclient.state());
    Serial.print("Attempting MQTT connection...");
    // Create the client ID
    String clientId = "SH-";
    clientId += String(s);
    // Attempt to connect
    if (mqttclient.connect(clientId.c_str())) {
      Serial.println("connected");
      Serial.print("Subscribe");
      mqttclient.subscribe (topicDO1);
      Serial.println(", Done");

      Serial.print("MQTT Loop");
      mqttclient.loop();
      Serial.println(", Done");
      Serial.print("State:");
      Serial.println(mqttclient.state());
    } else {
      Serial.print("failed, rc=");
      Serial.print(mqttclient.state());
      Serial.println(" try again in 5 seconds");
      // Wait 5 seconds before retrying
      for (uint16_t i = 0;i<500;i++){
        delay(10);
      }
      
    }
  }
  Serial.print("F: done in ");
  Serial.print(millis()-start_time);
  Serial.println(" mS");
```

When this code is run, the `mqttclient.loop();` it takes around 15 seconds to complete.
Previously this function did NOT have the mqttclient.loop() and was relying on the mqttclient.loop() inside the main program loop. I added it here as it was also taking 15 seconds in the main loop the first time the loop ran. As the mqttclient.loop() is blocking, at least performing the first call after an mqttclient.connect I have a little better control over it.

It is typically between 15100mS - 15800mS, however sometimes I have seen it as high as 40000mS.
Any ideas why this takes 15+ seconds?

gjt211:
I have found that changing two lines in PubSubClient.h has varying effects as follows...
If I change both MQTT_KEEPALIVE and MQTT_SOCKET_TIMEOUT to 20, then the first time I run mqttclient.loop() it takes a little over 20 seconds. (Remember sometimes, maybe 1 in 20 or more it connects within a few hundred milliseconds).

Next I tried with KEEPALIVE 10, and SOCKET_TIMEOUT 5.
And it connected in 5441mS.

[...]

PR #314, opened 2017-08-02
Fix issue #308

jorgecis:
This fix the issue #308 in esp8266 not sure for other boards, we need test it more.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -208,11 +208,8 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
 // reads a byte into result
 boolean PubSubClient::readByte(uint8_t * result) {
    uint32_t previousMillis = millis();
-   while(!_client->available()) {
-     uint32_t currentMillis = millis();
-     if(currentMillis - previousMillis >= ((int32_t) MQTT_SOCKET_TIMEOUT * 1000)){
+   if(!_client->available()) {
        return false;
-     }
    }
    *result = _client->read();
    return true;


knolleary:
Hi @jorgecis 

the code you've removed existed for a reason. It handles the case where there is a slow connection between the board and the broker, and the response from the broker don't arrive instantly. If you're only testing this on a local broker you won't hit the problem.

What could be happening here is that the tight loop in `readByte` is hammering the hardware, not giving it a chance to do what it needs to do - we have seen a similar thing it `client.loop()` is called too quickly.

Instead of your proposed fix, can you add a `delay(50)` inside the while loop? If that doesn't work, try increasing the delay. We wouldn't want to add too much of an artificial delay to the loop, but it would help understand the problem if it were to fix it.

jorgecis:
Hi @knolleary 
  

    My MQTT server is in the cloud, but is true the connection is via wifi and is fast. I didn't see any issue removing the code, and actually now is more fast. By the way I using TLS.

    I don't want add more delay, in the contrary I want remove the delay, f the problem right now is that in the first loop, the function  _client->available() return 5, and the code reads the 5 bytes, then the problem is when the function return 0, at this point they enter in the loop causing the 15 seconds delay ( until timeout), But this is not really necessary, we already read all bytes available.

    I saw your point. I think that we need found a better solution.

knolleary:
Please can you try adding a `delay(50)` inside the while loop - that will tell us if the underlying problem is because we're calling `available` too quickly in that loop. I'm not suggesting that will be the final fix, but it will at least allow us to understand the problem better.

A 50ms delay to fix a 15 second delay doesn't feel unreasonable to me.

stefanbode:
Hi, Knolleary. I have tried the change and also played around with the "fix". A delay before and/or after the .loop() request did not make any change to the behavior to let him run in the timeout. 
The change is doing the JOB very well, but I must admit that my MQQT server response quite fast. And I can confirm, that the timeout does not happen ALL the time. Quite often but not always. In my example, I send a 2 mqqt_publich and  3 mqqt_subscribe and then call the .loop() function during startup. Sub and pub go to different locations.

knolleary:
@stefanbode I'm not clear which change/fix you've tried. There's the change proposed in the original PR that I'm not in favour of as it assumes a good network.

There is then my suggested alternative to simply put a delay in the while loop this pr otherwise removes. (This is not the same as putting a delay in your own code around your calls to client.loop)

Can you clarify exactly what you've tested?

stefanbode:
I also tried the delay(50) in the while-loop just after the start. It looks like also this solves the problem. This would hint to your "hammering" idea as a root cause. I finally just inserted a yield(); and wow this is doing the same without a large time penalty

Issue #331, opened 2017-09-18
Functional MQTT_CALLBACK_SIGNATURE for ESP32


jaecktec:
Is there a reason to not have the functional feature on the esp32?
Ref to line: https://github.com/knolleary/pubsubclient/blob/master/src/PubSubClient.h#L78

maybe change it to 
`#if  defined(ESP8266) || defined(ESP32)`

knolleary:
I've never used the `ESP32` so cannot say if that makes sense or not. If it is valid - and you are able to test and verify the change, then a pull-request to add it would be welcome.

jaecktec:
OK already created the change in my branch. I'll create a pull request as soon as I can verify its actually running on the Hardware. 

In my case it makes sense because I wanted to pass a member function as a callback. This works with the functional interface but not with the old one.

PR #336, opened 2017-09-22
Enable ESP32 to use PubSubClient (make MQTT_CALLBACK_SIGNATURE functional)

jaecktec:
#331 

Added functional MQTT_CALLBACK_SIGNATURE to enable binding a object method to the callback handler.


sidoh:
Can confirm that this works nicely on an ESP32.  Thanks @jaecktec!

PR #360, opened 2017-11-16
Fix error reporting success status with longer payloads

eykamp:
Packet length is incorrectly calculated, which results in a failure status being returned when sending longer payloads that were, in fact, sent successfully.


knolleary:
Thanks - looks like this was fixed by another PR.

PR #361, opened 2017-11-16
Add some comments

eykamp:
This one is pretty lame.  But should be easy to review!  Might save someone a few seconds in understanding the code the first time they look at it.


knolleary:
Every little helps - thanks.

PR #362, opened 2017-11-16
Create new signature to permit cleaner user code

eykamp:
Allows user to write slightly simpler code:

// In the user code...
String json = "some payload";

bool ok = pubSubClient.publish_P("topic", json.c_str(), 0);     // Use new sig
vs.
bool ok = pubSubClient.publish_P("topic", (uint8_t*)json.c_str(), json.length(), 0);     // Use existing sig

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -376,6 +376,10 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
     return false;
 }
 
+boolean PubSubClient::publish_P(const char* topic, const char* payload, boolean retained) {
+    return publish_P(topic, (const uint8_t*)payload, strlen(payload), retained);
+}
+
 boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsigned int plength, boolean retained) {
     uint8_t llen = 0;
     uint8_t digit;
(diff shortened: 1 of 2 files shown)


knolleary:
Hi - thanks for all the PRs. I would prefer if you would raise an issue in the first instance so we can discuss specific changes before you start writing code.

Also, these changes would all need updates to the tests to ensure good coverage is maintained.

eykamp:
The PRs were all so minor it seemed discussing them first would have been too much.  Consider them proposals (except the bugfix).  Accept them or reject them as you see fit, though I'm happy to discuss further if you want.

I'll look at the test suite and see if I can craft a test to validate the bugfix; it failed consistently on longer packets -- I'm guessing the threshold is 255 bytes -- so adding a test with a longer payload should be trivial.

knolleary:
PRs for bug fixes are always welcome, and if they come with test coverage, all the better.

Proposing new APIs or functionality do warrant some conversation. It's up to you whether you want to spend time implementing a proposal before it's been discussed that could lead to changes.

But thanks for your contributions. I need to find some spare time to get them reviewed and merged. May be a little while.

eykamp:
I generally agree.  In this case, the time spent was minimal.  Most of the investment was understanding how the lib worked, which I had to do anyway to get it to work with my code.

I've written a test for sending long messages, but the test suite is not working for me.  Is it Python 2.7?

PR #365, opened 2017-12-07
Adding some speed improvments

czaraugust:
To put into practice some knowledge acquired in the programming for embedded systems subject, I've made some simple modifications to your code. 
I've changed divisions, modules, and multiplications for bitwise operations that are faster and less costly.
I got 4.86% of improvement in the use of the publish function.
I hope I have contributed and that the pull request is accepted, which would help me a lot in my s.ubject
Thanks in advance.


knolleary:
Thanks for these changes. My apologies it has taken so long for me to consider them.

src/PubSubClient.cpp
```diff
@@ -243,7 +244,7 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
         if(!readByte(&digit)) return 0;
         buffer[len++] = digit;
         length += (digit & 127) * multiplier;
-        multiplier *= 128;
+        multiplier <<=7; //multiplier *= 128
```
eku: The compiler's optimizer does exactly this. No need to do it manually.

src/PubSubClient.cpp
```diff
@@ -399,8 +400,8 @@ boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsig
     buffer[pos++] = header;
     len = plength + 2 + tlen;
     do {
-        digit = len % 128;
-        len = len / 128;
+        digit = len  & 127; //digit = len %128
+        len >>= 7; //len = len / 128
```
eku: The compiler's optimizer does exactly this. No need to do it manually.

src/PubSubClient.cpp
```diff
@@ -429,8 +430,9 @@ boolean PubSubClient::write(uint8_t header, uint8_t* buf, uint16_t length) {
     uint16_t rc;
     uint16_t len = length;
     do {
-        digit = len % 128;
-        len = len / 128;
+
+        digit = len  & 127; //digit = len %128
+        len >>= 7; //len = len / 128
```
eku: The compiler's optimizer does exactly this. No need to do it manually.

eku:
@knolleary Unfortunately my comment comes too late. The PR is already merged. Nevertheless, it should be noted that the changes do not bring any runtime advantage, but worsen the readability of the code.

knolleary:
Fair enough. At least the comments were left in place to keep the intention of the code clear.

PR #423, opened 2018-03-31
Enable PubSubClient for ESP32

twisniew1976:
This simple change will enable PubSubClient to work with ESP32 projects

### DIFF
--- a/src/PubSubClient.h
+++ b/src/PubSubClient.h
@@ -76,6 +76,9 @@
 #ifdef ESP8266
 #include <functional>
 #define MQTT_CALLBACK_SIGNATURE std::function<void(char*, uint8_t*, unsigned int)> callback
+#elif defined(ESP32)		// ESP32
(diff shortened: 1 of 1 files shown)


src/PubSubClient.h
```diff
@@ -76,6 +76,9 @@
 #ifdef ESP8266
```
domschl: Maybe:

`#if defined(ESP8266) or defined(ESP32)
`

to avoid dups

twisniew1976: I'm not sure that or will work, but:
#if defined(ESP8266) || defined(ESP32) 

is working

jaecktec: https://github.com/knolleary/pubsubclient/pull/336

already there :)

knolleary:
Merged #336 - so this PR isn't needed. Thanks anyway!

PR #494, opened 2018-09-04
Add large message API

### LINKED ISSUE #491: Add API to allow publishing of arbitrarily long messages
Having to modify the library to publish big messages is annoying, and having to keep a complete copy of the message in memory before it's sent also chews through a fair bit of memory.

We could add a new publish API, based on the UDP API in the core Arduino one, which has start and end functions, with any number of writes in between.

This might help mitigate issues #103 and #405 (I did think of commenting on those but they don't quite seem to match this)

amcewen:
This PR adds a new API to allow splitting publishing a message into three parts:
 1. Kicking off publishing, giving the topic and final message size
 1. Any number of calls to `print` or `write` to send an arbitrarily large payload
 1. A "message finished" call (mostly to mirror the similar APIs in the Arduino UDP and HttpClient APIs)

There's an example sketch which implements an "X green bottles" lyric generator, to show it in use.  This should fix #491.

### DIFF
--- a/examples/mqtt_large_message/mqtt_large_message.ino
+++ b/examples/mqtt_large_message/mqtt_large_message.ino
@@ -1,0 +1,179 @@
+/*
+ Long message ESP8266 MQTT example
+
+ This sketch demonstrates sending arbitrarily large messages in combination
+ with the ESP8266 board/library.
+
+ It connects to an MQTT server then:
+  - publishes "hello world" to the topic "outTopic"
+  - subscribes to the topic "greenBottles/#", printing out any messages
+    it receives. NB - it assumes the received payloads are strings not binary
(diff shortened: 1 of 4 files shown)


jksemple:
It seems cumbersome to have to calculate the message length before calling beginPublish.  Is it not possible to maintain a count internally of the bytes written/printed in the same way that is done in EthernetUDP::write using a _offset member variable? Then use that to set the message length in endPublish?

knolleary:
@jksemple no, because the length has to be written in the header of the MQTT packet, before the payload. The purpose of this API is to write the packet to the network with each call to write without buffering it in memory.

jksemple:
Gotcha. So, is there a way to use the richness of the parent Print class (e.g. to print PROGMEM strings) to write pieces of message content that could reasonably all be sent in a single packet and hence use the exising packet buffer?

PR #495, opened 2018-09-07
Update PubSubClient.cpp

flok99:
Make sure all data is flushed to the other end when doing a disconnect(): that way we know for sure that it is there when we disconnect the wifi or maybe even reboot.
This change was made after I noticed that I did not get any mqtt messages. I verified that it indeed solves the problem. Example code on request.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -524,6 +524,7 @@ void PubSubClient::disconnect() {
     buffer[1] = 0;
     _client->write(buffer,2);
     _state = MQTT_DISCONNECTED;
+    _client->flush();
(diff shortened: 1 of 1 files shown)


src/PubSubClient.cpp
```diff
@@ -524,6 +524,7 @@ void PubSubClient::disconnect() {
     buffer[1] = 0;
     _client->write(buffer,2);
```
TD-er: I noticed the `_client` pointer is never checked before dereferencing.
This makes it tricky to call the disconnect before switching off the wifi or go back to sleep, since you have to keep track outside of this library if the client ever has been set.

PR #528, opened 2018-11-21
Fix publish() when payload == NULL

lexszero:
Messages with empty payload are completely valid and are actually useful (e.g., to delete retained value).
This adds a check before calling `strlen()` to prevent crash when using shortcut methods taking `const char*`


knolleary:
Good spot - thanks.

PR #567, opened 2019-01-26
Add yield() calls in connect() and write() to avoid wdt resets if eit…

sprior:
…her blocks for too long


sprior:
This never got merged?

PR #568, opened 2019-01-28
Allow streaming of large messages

turgu1:
These changes are required to allow for the transmission of large messages through a connected stream. The changes do not have an impact on the class interface and habitual behavior. In particular, it will enable the use of OTA through a stream hooked through the setStream() class method. I've designed such a stream to demonstrate the functionality: https://github.com/turgu1/mqtt_ota_example.git

The change allows readPacket to get very large messages. At the moment, it is limited to 64Kb messages. Transmitting larger messages would allow to send firmware binary updates through the mosquitto_pub command or any other mean as one single message. The change is limited to the readPacket method. The running length variable is transformed to be an int32 instead of an int16.

Guy

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -197,7 +197,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
                 }
             }
             uint8_t llen;
-            uint16_t len = readPacket(&llen);
+            uint32_t len = readPacket(&llen);
 
             if (len == 4) {
                 if (buffer[3] == 0) {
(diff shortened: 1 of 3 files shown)


knolleary:
Hi, could you explain a bit more about why the pr is necessary and what changes actually do?

You may not have noticed, but the library includes unit tests that this change has broken.

turgu1:
My call!! sorry about that, I've omitted the one line change in the .h file such that the tests would pass...

turgu1:
After my corrective action, testing failed again. The current fail is due to the change that allow for big messages. I thing this is normal because of the change I made.

turgu1:
Hello. I modified the test case on oversized messages to get it pass.

knolleary:
Thanks for the fix - it's appreciated.

Issue #220, opened 2016-12-19
boolean PubSubClient::publish_P returns false 


andrewkpearce:
We think there is a bug in the publish_P method which was reporting false, even though the message was sent to the server. The expected length is off by a few bytes.

The original code:

     return rc == tlen + 4 + plength;              

Was changed to:

    // Header (1 byte) + llen + identifier (2 bytes)  + topic len + payload len                                                                               
    int expectedLength = 1 + llen + 2 + tlen + plength;
    return (rc == expectedLength);                                                                                     
}

Lucianovici:
Any progress on this? Nobody wants to fix this?

PR #600, opened 2019-04-07
Fix publish_P return value

cordu-dev:
Fix for https://github.com/knolleary/pubsubclient/issues/220

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -412,6 +412,7 @@ boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsig
     unsigned int i;
     uint8_t header;
     unsigned int len;
+    int expectedLength;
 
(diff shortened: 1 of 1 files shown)


knolleary:
Ah yes - variable length encoding of the length field. Thanks for providing the fix!

src/PubSubClient.cpp
```diff
@@ -412,6 +412,7 @@ boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsig
     unsigned int i;
     uint8_t header;
     unsigned int len;
+    int expectedLength;
```
smartynov, line 415: This should probably be an `unsigned int` (avoid confusion and compiler warning)

hackersboss36436:
__unsigned int i;
    uint8_t header;
    unsigned int len;

PR #622, opened 2019-05-23
Changing index and length to uint32_t to be able to handle messages w…

Romkabouter:
…ith a length > 65535

For a project I needed to be able to receive wave files in a single payload buffer.
I ran into a crash whenever the payload length was larger than around uint16_t max value.
I have changed this on several places to uint32_t and the handling has been running fine since.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -233,8 +233,8 @@ boolean PubSubClient::readByte(uint8_t * result) {
 }
 
 // reads a byte into result[*index] and increments index
-boolean PubSubClient::readByte(uint8_t * result, uint16_t * index){
-  uint16_t current_index = *index;
+boolean PubSubClient::readByte(uint8_t * result, uint32_t * index){
+  uint32_t current_index = *index;
   uint8_t * write_address = &(result[current_index]);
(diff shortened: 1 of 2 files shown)


Romkabouter:
I do not understand why the check says it hasn't completed. When I click Details I see all green?

knolleary:
Probably just a hiccough when Travis tried to report status back to github.

All looks okay to me.

knolleary:
Looking at your changes, you have changed the return type of `readPacket` but there are two places in the code where `readPacket` is called and the returned value treated as `uint16_t` - one in `connect` and one in `loop`.

That length value is also returned (after some calculation) as part of the message callback function signature. That signature defines its type as `unsigned int`. I have to admit, my arduino usage is very rusty these days. Is passing a `uint32_t` back as `unsigned int` safe to do?

Romkabouter:
Hmm, yes that might be a problem on 16bit systems so I am not sure unfortunatly

hutje:
@Romkabouter did you make it work with uint32_t? I also need more package size...

Romkabouter:
@hutje no, I am using https://github.com/marvinroger/async-mqtt-client

PR #628, opened 2019-06-04
update lastOutActivity if puback is successful

a-c-sreedhar-reddy:


knolleary:
Is this a result of a specific issue you have hit or just through inspecting the code?

a-c-sreedhar-reddy:
Just through inspecting the code.

TD-er:
There are more occurrences of `client->write`.
Why don't they need the same check?

knolleary:
If you change `lastOutActivity` at this point, then the logic in https://github.com/knolleary/pubsubclient/blob/dd3a48a86d4d899936158ab6bb1d03038be29cc2/src/PubSubClient.cpp#L305 will not work and we'll never spot a ping is outstanding.

TD-er:
@knolleary Can you elaborate on that?
As a test in my local copy of your library (with the intent to make a PR once it appears to be stable and has been thoroughly tested) I did something similar. Only I did it on multiple occasions where `_client->write` is called.
See https://github.com/letscontrolit/ESPEasy/blob/mega/lib/pubsubclient/src/PubSubClient.cpp (ctrl-f on `lastOutActivity ` )

a-c-sreedhar-reddy:
Then I think these lines should be moved into if condition
https://github.com/knolleary/pubsubclient/blob/master/src/PubSubClient.cpp#L323-L325

knolleary:
@TD-er I take it back - you're right.

Although it isn't necessary to update `lastOutActivity` after every single call to `client->write`. There are a few places where a call will lead to multiple `client->write` calls, so `lastOutActivity` only needs to be updates once after the batch of calls. There's no harm updating it in all of those extra places... just may be more than is needed.


@a-c-sreedhar-reddy you'll have to expand on that... which if condition do you mean? What issue does it solve?

TD-er:
@knolleary 
Thanks for getting back on this :)

My intent for these changes (was triggered to look into it after reading this issue, so credits to @a-c-sreedhar-reddy ) was to address the following:
If the write fails, it means the number of bytes written was 0, or the client object timed out, or we have no WiFi connection. 
So from the broker's point of view, there has been no activity and thus we must not renew the last out activity timer.
This timer is used to determine when to send a keep-alive thingy.

The actual problems I do see, is that with bad WiFi link quality, the MQTT client does perform a lot of reconnects.   The number of reconnect issues is actually reduced by the changes I have made.

Another thing I do see (and which is sadly not fixed by this change) is that the node does run out of memory with MQTT active and bad WiFi link quality.
Not sure if that's part of your library (very likely not), or my own code, or the esp8266/Arduino core.
But with the changes I made, the memory exhaustion is less than before, so it looks like it is somewhat related to the reconnect.

a-c-sreedhar-reddy:
@knolleary if `client->write` method returns 0 bytes it means there is no `lastOutActivity` and hence it should not be updated.

Issue #583, opened 2019-03-09
PubSubClient::connect is blocking the main loop :(


gibman:
hi..

Today I disovered that the call to PubSubClient::connect is blocking as my MQTT was down.

Each failed call to connect is taking 5sec.
obvioulsy when the MQTT broker is running all is nice and behaving. But I want it to handle these unfortunate circumstances as well.

This means that if I have other time critical stuff going on in the main loop then Im out of luck.

Any ideas ?

knolleary:
Connect is a blocking call. Sorry.

There is already an issue asking for nonblocking connect: https://github.com/knolleary/pubsubclient/issues/579

Not much else I can suggest.

Domochip:
An intermediate solution would be to change the Timout of your Client to reduce this time of 5sec.
ex : myWiFiClient.setTimeout(1000);

mlewand:
@knolleary I believe that the first issue requesting for non-blocking connect feature is #147. I think it make sense to close this issue and #579 as duplicates of #147 and focus the discussion there.

spatnynick:
It seems I successfully solved the blocking by using hw_timer and/or tickers, both is working. The main loop is blocked, I have simply moved my logic there
https://circuits4you.com/2018/01/02/esp8266-timer-ticker-example/

mlewand:
@spatnynick would you like to share the code?

spatnynick:
Ok there is the code

Its a MOSFET LED dimmer for my strips in aquarium, controlled only via MQTT. Timing is fine-tuned for Wemos D1 mini. With other boards I guess this timing might need to be adjusted

I use hw_timer to dimm LEDs smoothly -> this processing should have constant runtime so the LED are not blinking
In the ticker routine I do network communication with MQTT

All is green, up&running for a week now? ;)


`#include <ESP8266WiFi.h>
#include <Ticker.h>  //Ticker Library
#include <PubSubClient.h>

#define TOPIC_CMD_LED_INTENSITY "ESPdimmer01/LEDintensity"
#define TOPIC_STAT_LED_INTENSITY "ESPdimmer01/stat/LEDintensity"

#define LED 5  // wemos D1
//#define LED 2 // esp01

#define STEP_TIME 455          // fine-tune this value; lower = dimmer less blinks
#define FADE_STEP 5;           // fine-tune this value; lower = slower diming (intensity change)

Ticker tickMaintain;

String ssid = <WIFI_SSID>;
[...]

Issue #579, opened 2019-03-06
Asynchronous connect


9a4gl:
Can we have asynchronous connect ? I need to update display, own watchdog and other stuff during connecting. Or at least a connect to which we provide own yield function that will be called.

atuline:
My outdoor LED displays halt and/or stutter when my MQTT server (my phone) goes in and out of range. Have yet to find an MQTT library that doesn't block the loop in some manner. As a result, I've had to remove MQTT (or pubsubclient) from the majority of those displays.

WebDust21:
I ran up against this exact same problem.  Tried two solutions...
...for the first, I gutted the "connect" routine, putting the blocking "are we connected yet?" into the LOOP call, as well as adding a new MQTT state handler for "connecting" mode.  You might notice that LOOP only does anything if the socket is connected...which makes it very easy to add code for a non-connected state.
I'd provide a lot more documentation EXCEPT that I can't say that it helped at all.  It didn't break anything, though.

Second solution?  Redesign multiple thousands of lines of my program code to use an RTOS setup...putting all the WiFi stuff on its own thread on the other core.  Problem solved.

atuline:
I moved on from pubsubclient to the async-mqtt-client library and am now onto WLED. Am learning it.

ilker07:
#include <WiFi.h>
extern "C" {
  #include "freertos/FreeRTOS.h"
  #include "freertos/timers.h"
}
#include <AsyncMqttClient.h>

#define TINY_GSM_MODEM_SIM800 // Modem is SIM800L
#define SerialMon Serial
#define SerialAT Serial1
#define TINY_GSM_DEBUG SerialMon



const char apn[] ="internet";//internet
const char gprsUser[] ="";
const char gprsPass[] ="";

#include <Wire.h>
#include <TinyGsmClient.h>

#ifdef DUMP_AT_COMMANDS
  #include <StreamDebugger.h>
  StreamDebugger debugger(SerialAT, SerialMon);
  TinyGsm modem(debugger);
#else
  TinyGsm modem(SerialAT);
#endif

TinyGsmClient client(modem);

// Digital Ocean MQTT Mosquitto Broker
#define MQTT_HOST IPAddress(xxx,xxx,xxx,xxx)
#define MQTT_PORT 1883

#define MQTT_USERNAME "REPLACE_WITH_YOUR_MQTT_USER"
[...]

Issue #147, opened 2016-03-30
nonblocking approach is indeed blocking 


glebv:
I used example pubsubclient/examples/mqtt_reconnect_nonblocking/mqtt_reconnect_nonblocking.ino as base and added work with a button. Thus, If it calls `client.connect(...)` my script is unable to read button state in that moment. It's short period about 1-2s but it's weird. Is it possible to handle it in some way? 

There's example of important part of my app
https://gist.github.com/glebv/4b79298fc7ac23f342a42017cb77e6cd

knolleary:
The point of the example is that it doesn't block the app the entire time the client is disconnected - so if it loses its network connection for 5 minutes, the sketch can continue running.

But of course there is work involved in getting connected and that can take a little bit of time. The current API assumes connect returns true/false based on whether it got connected or not. If we changed it to be non-blocking, such that it would connect eventually, that would break a lot of sketches. An alternative would be to add an alternative connect function that wasn't blocking. But that isn't the quick solution you're looking for - I don't know when I'll get a chance to add such a function to the library.

glebv:
I see, I guess I will be very helpful! Am I right that only one chance to reduce the effect of the connection delay is to increase the time between the reconnections?

Findus76:
Hello
Have the same problem.
Due to some bug between WiFi or pub subclient, the device can be locked. It stops executing the loop routine. But it seems to continue kicking on the internal watchdog. In an attempt to solve this without having to dig into the WiFi or pubsubclient I try with an external watchdog. But then got a direct problem, those times that the pubsubclient should make connection and get timeout. Then the execution of the loop stops for a few seconds. The external watchdog that has about 1.2s of timeout jumps in and resets.
Should try version 6 soon, see if there remains the same problem (bug and timeout).
Using default settings for Timeout, 15s.

atuline:
While not perfect, it's the best workaround for the blocking effects of the library that I've seen. Anything that makes my ESP8266 based animated LED displays stutter is immediately noticeable and this library has indeed been a killer if I lose connection to the broker.

bobcroft:
The solution shown by Spatnynick looks helpful.  Are there things one can do to ensure the broker responds as fast as possible?  The delay associated with MQTT reconnect messed up an application where I was trying to use rs485 to achieve control in an area where WiFi wasn't available.  I wasn't successful in trying to control when the mqtt reconnect attempt happened in order to better control the rs485 side.

mattie47:
[...]

PR #640, opened 2019-07-16
Add non-blocking connect functions and ESP8266 example

brianrho:
First, to make the code easier to grapple with, the nesting levels in `connect()` and `loop()` were reduced with "early returns", without touching any of the actual logic. But I understand this is a matter of style, so it can always be removed if not desired.

Since non-blocking connects are a feature I need and have seen requested here in a few issues, I've created 2 functions for this: `beginConnect()` and `connectStatus()`. These 2 functions are basically copy-pastes of 2 separate parts of the existing `connect()` so that the functionality is still the same.

`connected()` was modified to check the `_state` as well. This was necessary for the new functions to work properly and I don't think this will break anything either, if anything, `connected()` should now be more precise about the connection's status. I tested with the `mqtt_esp8266` example, to be sure.

All other functions remain the same, so that nothing breaks even if the new functions don't work or have any bugs. The new functions do work just fine, as expected though. I tested with the new `mqtt_esp8266_nonblocking` example, breaking the connection and then re-connecting. Here's a screenshot of the serial output:

Screenshot from 2019-07-16 04-01-44

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -117,14 +117,16 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
     return connect(id,user,pass,willTopic,willQos,willRetain,willMessage,1);
 }
 
-boolean PubSubClient::connect(const char *id, const char *user, const char *pass, const char* willTopic, uint8_t willQos, boolean willRetain, const char* willMessage, boolean cleanSession) {
+boolean PubSubClient::connect(const char *id, const char *user, const char *pass, const char* willTopic, uint8_t willQos, boolean willRetain, const char* willMessage, boolean cleanSession) {  
     if (connected()) {
         return true;
     }
         
     int result = 0;
 
-    if (domain != NULL) {
+    if(_client->connected()) {
+        result = 1;
+    } else if (domain != NULL) {
         result = _client->connect(this->domain, this->port);
     } else {
         result = _client->connect(this->ip, this->port);
@@ -205,7 +207,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
         }
     }
     uint8_t llen;
-    uint16_t len = readPacket(&llen);
+    uint32_t len = readPacket(&llen);
 
     if (len == 4) {
         if (buffer[3] == 0) {
@@ -245,7 +247,9 @@ boolean PubSubClient::beginConnect(const char *id, const char *user, const char 
     int result = 0;
     
     _client->stop();
-    if (domain != NULL) {
+    if(_client->connected()) {
+        result = 1;
+    } else if (domain != NULL) {
         result = _client->connect(this->domain, this->port);
     } else {
         result = _client->connect(this->ip, this->port);
@@ -380,12 +384,12 @@ boolean PubSubClient::readByte(uint8_t * result, uint16_t * index){
   return false;
 }
 
-uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
+uint32_t PubSubClient::readPacket(uint8_t* lengthLength) {
     uint16_t len = 0;
     if(!readByte(buffer, &len)) return 0;
     bool isPublish = (buffer[0]&0xF0) == MQTTPUBLISH;
     uint32_t multiplier = 1;
-    uint16_t length = 0;
+    uint32_t length = 0;
     uint8_t digit = 0;
     uint16_t skip = 0;
     uint8_t start = 0;
@@ -416,20 +420,22 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
         }
     }
 
-    for (uint16_t i = start;i<length;i++) {
+    uint32_t idx = len;
+    for (uint32_t i = start;i<length;i++) {
         if(!readByte(&digit)) return 0;
         if (this->stream) {
-            if (isPublish && len-*lengthLength-2>skip) {
+            if (isPublish && idx-*lengthLength-2>skip) {
                 this->stream->write(digit);
             }
         }
         if (len < MQTT_MAX_PACKET_SIZE) {
             buffer[len] = digit;
+            len++;
         }
-        len++;
+        idx++;
     }
 
-    if (!this->stream && len > MQTT_MAX_PACKET_SIZE) {
+    if (!this->stream && idx > MQTT_MAX_PACKET_SIZE) {
         len = 0; // This will cause the packet to be ignored.
     }
 
@@ -511,11 +517,11 @@ boolean PubSubClient::loop() {
 }
 
 boolean PubSubClient::publish(const char* topic, const char* payload) {
-    return publish(topic,(const uint8_t*)payload,strlen(payload),false);
+    return publish(topic,(const uint8_t*)payload, payload ? strnlen(payload, MQTT_MAX_PACKET_SIZE) : 0,false);
 }
 
 boolean PubSubClient::publish(const char* topic, const char* payload, boolean retained) {
-    return publish(topic,(const uint8_t*)payload,strlen(payload),retained);
+    return publish(topic,(const uint8_t*)payload, payload ? strnlen(payload, MQTT_MAX_PACKET_SIZE) : 0,retained);
 }
 
 boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigned int plength) {
@@ -531,10 +537,14 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
         // Leave room in the buffer for header and variable length field
         uint16_t length = MQTT_MAX_HEADER_SIZE;
         length = writeString(topic,buffer,length);
+        
+        // Add payload
         uint16_t i;
         for (i=0;i<plength;i++) {
             buffer[length++] = payload[i];
         }
+        
+        // Write the header
         uint8_t header = MQTTPUBLISH;
         if (retained) {
             header |= 1;
@@ -545,7 +555,7 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
 }
 
 boolean PubSubClient::publish_P(const char* topic, const char* payload, boolean retained) {
-    return publish_P(topic, (const uint8_t*)payload, strlen(payload), retained);
+    return publish_P(topic, (const uint8_t*)payload, payload ? strnlen(payload, MQTT_MAX_PACKET_SIZE) : 0, retained);
 }
 
 boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsigned int plength, boolean retained) {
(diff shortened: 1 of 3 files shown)


MBerka (commented):
Specifically, there is an overload of `beginConnect()` corresponding to every existing `connect()`. Can confirm that all the original code is still there, and header/example look fine. Would like to add tests to cover the new capability if that's okay, and eventually document somewhere outside /examples. Would it maybe be clearer to users to name the new connect function `asyncConnect()` or `nonBlockingConnect()`?

brianrho:
I used `beginConnect()` for simplicity and in keeping with Arduino tradition, but `asyncConnect()` also works.

MBerka:
> I used `beginConnect()` for simplicity and in keeping with Arduino tradition, but `asyncConnect()` also works.

Makes sense, I found it on Ethernet and Bluetooth clients, among others. Never mind that part.

mattie47:
Unless I'm doing something quite stupid, I haven't been able to get this example working either with non-blocking.

I'm using VS code, and have specified to use the fork's library, and built the accompanying example.

However, any code I add into the main loop becomes blocked as soon as the esp loses mqtt connectivity.

Correct me if I'm wrong, but isn't the purpose of the PR to ensure the main loop can continue running while the ESP tries to reconnect to the broker in the background?

I just added a simple blink of the on board LED and this gets blocked for 5 seconds when broker connectivity is lost. 

I came across this PR after coming across behaviour observed in:
https://github.com/knolleary/pubsubclient/issues/147
https://github.com/knolleary/pubsubclient/issues/583
https://github.com/knolleary/pubsubclient/issues/579


Thanks,

Matt

brianrho:
@mattie47 Yes, the PR is supposed to ensure MQTT (re)connects don't block. However, the `mqtt_esp8266_nonblocking` example does not concern itself with fixing *Wi-Fi* (re)connections that may take a few seconds. So if your ESP lost not only MQTT connectivity but also access to the AP, then it may explain why your code is still blocking. This is easily fixed though, but let's see the code first to be sure that's the problem.

mattie47:
@brianrho Thanks for the reply. I'm not disconnecting WiFi. Only terminating the mqtt server daemon. I just added a simple blink of the led to the main loop in the example sketch (See below). So this works fine if the server is running, but as soon as I terminate the server, the led blinking stops and only resumes twice after 5 seconds:


```c
/*
  Basic ESP8266 MQTT example with non-blocking connect

  This sketch demonstrates the capabilities of the pubsub library in combination
  with the ESP8266 board/library.

  It initiates a non-blocking connection to an MQTT server then:
  - polls the connection status till it's completed (or times out), letting you perform
    other non-MQTT tasks in the meantime.
    
    Then it:
  - publishes "hello world" to the topic "outTopic" every two seconds
  - subscribes to the topic "inTopic", printing out any messages
    it receives. NB - it assumes the received payloads are strings not binary
  - If the first character of the topic "inTopic" is an 1, switch ON the ESP Led,
    else switch it off

  It will reconnect (also in a non-blocking manner) to the server if the connection is lost.

  To install the ESP8266 board, (using Arduino 1.6.4+):
  - Add the following 3rd party board manager under "File -> Preferences -> Additional Boards Manager URLs":
       http://arduino.esp8266.com/stable/package_esp8266com_index.json
  - Open the "Tools -> Board -> Board Manager" and click install for the ESP8266"
  - Select your ESP8266 in "Tools -> Board"

*/

#include <ESP8266WiFi.h>
#include <PubSubClient.h>

const char* ssid = "adsf";
const char* password = "sadf";
const char* mqtt_server = "192.168.1.47";

WiFiClient espClient;
PubSubClient client(espClient);
long lastMsg = 0;
char msg[50];
int value = 0;

void setup_wifi() {

  delay(10);
  // We start by connecting to a WiFi network
  Serial.println();
  Serial.print("Connecting to ");
  Serial.println(ssid);

  WiFi.begin(ssid, password);

  while (WiFi.status() != WL_CONNECTED) {
    delay(500);
    Serial.print(".");
  }

  randomSeed(micros());

  Serial.println("");
  Serial.println("WiFi connected");
  Serial.println("IP address: ");
  Serial.println(WiFi.localIP());
}

void callback(char* topic, byte* payload, unsigned int length) {
  Serial.print("Message arrived [");
  Serial.print(topic);
  Serial.print("] ");
  for (int i = 0; i < length; i++) {
    Serial.print((char)payload[i]);
  }
  Serial.println();

  // Switch on the LED if an 1 was received as first character
  if ((char)payload[0] == '1') {
    digitalWrite(BUILTIN_LED, LOW);   // Turn the LED on (Note that LOW is the voltage level
    // but actually the LED is on; this is because
    // it is active low on the ESP-01)
  } else {
    digitalWrite(BUILTIN_LED, HIGH);  // Turn the LED off by making the voltage HIGH
  }

}

long lastReconnectAttempt = 0;

void reconnect() {
  // check that a connection isnt already in progress
  if (client.state() != MQTT_CONNECT_INPROGRESS) {
    long now = millis();
    if (now - lastReconnectAttempt < 5000)
      // Serial.println("if statement true");
      // delay(500);
      return;

    Serial.print("Now: "); Serial.println(now);
    Serial.print("lastReconnectAttempt: "); Serial.println(lastReconnectAttempt);
    Serial.println("Attempting MQTT connection...");
    
    // Create a random client ID
    String clientId = "ESP8266Client-";
    clientId += String(random(0xffff), HEX);

    // Attempt to connect
    lastReconnectAttempt = now;
    client.beginConnect(clientId.c_str());
  }

  // check the connect status
  int ret = client.connectStatus();
  switch (ret) {
    case MQTT_CONNECTED:
      Serial.println("connected");
      // Once connected, publish an announcement...
      client.publish("outTopic", "hello world");
      // ... and resubscribe
      client.subscribe("inTopic");

      lastReconnectAttempt = 0;
      break;
    case MQTT_CONNECT_INPROGRESS:
      return;
    default:
      Serial.print("failed! rc = "); Serial.print(ret);
      Serial.println(". Trying again in 5 seconds.");
      break;
  }
}

void setup() {
  pinMode(BUILTIN_LED, OUTPUT);     // Initialize the BUILTIN_LED pin as an output
  Serial.begin(115200);
  setup_wifi();
  client.setServer(mqtt_server, 1883);
  client.setCallback(callback);
}

void loop() {
    // LED: LOW = on, HIGH = off
  Serial.println("Start blinking");
  for (int i = 0; i < 2; i++)
  {
    digitalWrite(BUILTIN_LED, LOW);
    delay(100);
    digitalWrite(BUILTIN_LED, HIGH);
    delay(100);
  }

  if (!client.connected()) {
    reconnect();
  }
  else {
    client.loop();

    long now = millis();
    if (now - lastMsg > 2000) {
      lastMsg = now;
      ++value;
      snprintf (msg, 50, "hello world #%ld", value);
      Serial.print("Publish message: ");
      Serial.println(msg);
      client.publish("outTopic", msg);
    }
  }
}
```

brianrho:
So I tried this with a NodeMCU and a local mosquitto broker and found:

- The sketch works fine so long as the mosquitto broker service is simply stopped or killed. It doesn't delay during reconnection at all, even if no server is listening.

- If the broker port is also blocked by a firewall however, the client spends 5 secs waiting to connect, specifically during `client.connect()` in the `beginConnect()` method. 

It seems in the first case, a TCP RST packet is returned immediately by the server's TCP stack so the client is able to conclude immediately that the connection has been reset/refused. But in the second case where the port is "filtered", the firewall simply drops the packets and doesn't respond at all (for security purposes), so the client waits for a response in vain for 5 seconds before returning, the default in `ESP8266WiFi.h`. I'm assuming this is what happened in your case somehow: there was no response at all from the server for some reason.

You've got two options:
- Write/find a port of this library that uses the [ESPAsyncTCP](https://github.com/me-no-dev/ESPAsyncTCP) client library instead of the default client.
- Use `espClient.setTimeout()` to reduce the connect wait interval as you like. I tried it to 2 seconds to be sure that it works.

vladosam:
Thanks for trying to fix this. If this PR gets merged will it work on arduino?

brianrho:
Yeah, it will. The proposed changes are not device-specific, though the example provided is for an ESP8266 because its WiFi allows for easy testing.

vladosam:
Now we just have to wait. :smile: :hourglass_flowing_sand:

mattie47:
> So I tried this with a NodeMCU and a local mosquitto broker and found:
> 
> * The sketch works fine so long as the mosquitto broker service is simply stopped or killed. It doesn't delay during reconnection at all, even if no server is listening.
> * If the broker port is also blocked by a firewall however, the client spends 5 secs waiting to connect, specifically during `client.connect()` in the `beginConnect()` method.
> 
> It seems in the first case, a TCP RST packet is returned immediately by the server's TCP stack so the client is able to conclude immediately that the connection has been reset/refused. But in the second case where the port is "filtered", the firewall simply drops the packets and doesn't respond at all (for security purposes), so the client waits for a response in vain for 5 seconds before returning, the default in `WiFiClient.cpp`. I'm assuming this is what happened in your case somehow: there was no response at all from the server for some reason.
> 
> You've got two options:
> 
> * Write/find a port of this library that uses the [ESPAsyncTCP](https://github.com/me-no-dev/ESPAsyncTCP) client library instead of the default client.
> * Use `espClient.setTimeout()` to reduce the connect wait interval as you like. I tried it with 2 seconds to be sure that it works.


Thanks for the testing and response.

My issue is that I'm using an ESP8266 with an LCD screen and if the MQTT server is not available, it would freeze up the screen. I couldn't easily find a way around the blocking architecture. So `espClient.setTimeout()` wouldn't have helped in my case. 

I probably could have gone to the effort of implementing ESPAsyncTCP, and I thank you for pointing this out. 

Instead I changed from pubsubclient to [AsyncMqttClient](https://github.com/marvinroger/async-mqtt-client) which has worked really well with my testing :-).

brianrho:
Glad you found something that worked. Seems that library uses `ESPAsyncTCP` under the hood. Regarding the LCD though, it's still possible to use this so long as you aren't refreshing the display too often and your deadlines are flexible. The client timeout just needs to be set to be well under the refresh interval, but still long enough that the client has enough time to wait for a response to its TCP connect gesture. Probably not very useful for refresh intervals less than a few seconds.

CircuitSetup:
Did anyone get this working properly?

brianrho:
@CircuitSetup Yeah, it's working alright, just never got merged into this branch.

src/PubSubClient.cpp
```diff
@@ -416,20 +420,22 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
         }
     }
 
-    for (uint16_t i = start;i<length;i++) {
+    uint32_t idx = len;
+    for (uint32_t i = start;i<length;i++) {
         if(!readByte(&digit)) return 0;
         if (this->stream) {
-            if (isPublish && len-*lengthLength-2>skip) {
+            if (isPublish && idx-*lengthLength-2>skip) {
```
TD-er, line 427: I know it wasn't in the original code, but shouldn't here be some kind of braces be added? At least for readability.
Also the part right from the && is now using 32, 8 and 16 bit values.
A `static_cast` would be a good indication someone at least thought about it, but I guess the `skip` should also be a 32 bit value if you also switch to 32 bit for the max. message length.  (not sure if that's possible with MQTT though)

brianrho: Those are okay changes but I think they're pretty minor and should be made onto the main branch, with the non-blocking changes from this branch manually merged into that as well, if found worthy. It's been too long since I submitted this PR, there's apparently a bunch of conflicts now and who knows how many tests failing. Spent too much time on this already, trying to sync up.

PR #651, opened 2019-08-15
Fixed Bad Buffer SHORT MQTT Payload

ajaybnl:
I used publish with "1" short value, when i see the subscribed output in serial, we see random chars with the values we sent using MQTT.fx from windows, Android App.
in loop:
added:
 for(int a=0;a<MQTT_MAX_PACKET_SIZE;a++){  buffer[a]=0;   }

The FIX cleans the buffer before receiving. Now, the output of subscription payload is neat.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -300,7 +300,8 @@ uint16_t PubSubClient::readPacket(uint8_t* lengthLength) {
 }
 
 boolean PubSubClient::loop() {
-    if (connected()) {
+    if (connected()) {   
+   
         unsigned long t = millis();
         if ((t - lastInActivity > MQTT_KEEPALIVE*1000UL) || (t - lastOutActivity > MQTT_KEEPALIVE*1000UL)) {
             if (pingOutstanding) {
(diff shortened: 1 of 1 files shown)


knolleary:
The callback function provides the length of the received message. You should not be reading more bytes then that. This avoids the overhead of clearing the buffer each time.

ajaybnl:
I have publish + subscription sketch where i observe the problem. Please check sending large and small numbers from android app slider from 1 to 99999.

knolleary:
Please share the code of your callback handler that sees the problem.

ajaybnl:
It is just serial.println( (char*)payload) and topic.

knolleary:
> It is just serial.println( (char*)payload) and topic.

Then your code is not using the callback arguments properly. The third argument of the callback tells you how long the payload is. You should make use of that. For example, you could null-terminate the payload with:
```
payload[length] = "\0";
```

And then you can treat it as a null-terminated string as you are.

ajaybnl:
Did you changed your Example to use payload[length] = '\0'; because it works.

PR #653, opened 2019-08-16
payload should be exactly what we send. Not the buffer.

ajaybnl:
Payload is a buffer, if you don't trim it, it is a chunk that start with your data + other messages leftovers.

so i added in callback to get only the payload:

 //payload extract
 payload[length] = '\0';

### DIFF
--- a/examples/mqtt_esp8266/mqtt_esp8266.ino
+++ b/examples/mqtt_esp8266/mqtt_esp8266.ino
@@ -62,6 +62,9 @@ void setup_wifi() {
 }
 
 void callback(char* topic, byte* payload, unsigned int length) {
+ //payload extract
+ payload[length] = '\0';
+ 
   Serial.print("Message arrived [");
   Serial.print(topic);
   Serial.print("] ");


abderraouf-adjal:
Hello,

IIRC this modification will cause a data loss from payload, for example when a `1` byte string sent such as `"P"` (which is `{'P', '\0'}`) the `strlen()` return length will be `1`, not `2`, so there will be no data passed to process since `'P'` will become `0x00`.

If I'm wrong did you do unit testing for this?

ajaybnl:
I have tested it using "1" and "0" and they work fine. Also the length will be 1 but the buffer starts from 0 so the single byte will be preserved.

knolleary:
@abderraouf-adjal if you send `{'P', '\0'}` then you are sending a 2 byte payload. The `length` parameter passed to the callback will be `2`. So doing `payload[length] = '\0'` will result in `{'P', '\0', '\0'}` and all is well.

abderraouf-adjal:
> @abderraouf-adjal if you send `{'P', '\0'}` then you are sending a 2 byte payload. The `length` parameter passed to the callback will be `2`. So doing `payload[length] = '\0'` will result in `{'P', '\0', '\0'}` and all is well.

Interesting, in my current project I use Adafruit IO server and [this app](https://play.google.com/store/apps/details?id=snr.lab.iotmqttpanel.prod), I send one character from the app and in esp8266 I check the length, it's always 1.

abderraouf-adjal:
@ajaybnl 
I'm not sure if there is absolutely no risk by doing this because lengths calculation is a bit confusing,
See [the buffer declaration](https://github.com/knolleary/pubsubclient/blob/d2ff771f4a79950fa36e35c4c71e232da4f85f5d/src/PubSubClient.h#L91) and [this part of code](https://github.com/knolleary/pubsubclient/blob/d2ff771f4a79950fa36e35c4c71e232da4f85f5d/src/PubSubClient.cpp#L290).
My question is: Can this modification cause data loss if the packet filled all that buffer?
I also need to note that there is no need for this since we are printing [char by char here](https://github.com/knolleary/pubsubclient/blob/91ad6f67d2205ed4f861829bbf864e2c787e3753/examples/mqtt_esp8266/mqtt_esp8266.ino#L72), so maybe pointing by a comment that the payload isn't null-terminated array will be better.

knolleary:
No, this change cannot corrupt it.

The length argument to the callback tells you exactly how many bytes of payload to read. The one extreme edge case is if the received packet fills the entire buffer, in which case this will overrun the buffer.

For that reason, I'm not going to be merging this change into the example code. Individual users need to decide for themselves of that is suitable given their application and circumstances.

ajaybnl:
Dear knolleary, Like i had problems using strcmp to match payload, if you want somebody not having problems, please add a comment to have the line i wrote, so that he should use it if needed.

I have been troubled for 2 hours figuring it out, and for the easiness of new arduino coders,  please do it.

ajaybnl:
abderraouf-adjal you are considering \0 as a char, but its a termination symbol for arduino. It should send your "P" in your callback message.

uahrendt:
Hi,
I think, there is another problem with using 
`payload[length] = '\0';`

The buffer and its size is defined via
`#define MQTT_MAX_PACKET_SIZE 128`
and
`uint8_t buffer[MQTT_MAX_PACKET_SIZE];`

So in this case the buffer reaches from `buffer[0]` to `buffer[127]`.

The data written to the buffer is limited to the max size:
`if (len < MQTT_MAX_PACKET_SIZE)`
`{`
` buffer[len] = digit;`
`}`

So the last possible value is stored in `buffer[127]`. 

If you send a 128 character payload the length parameter will have the (correct) value 128, the data is in` buffer[0]` to `buffer[127]`.

So far so good.

But:
With
`payload[length] = '\0';`
you write a character to buffer[128] with is not defined/allowed! You are overwriting the next byte after `buffer` in memory (buffer overflow).

abderraouf-adjal:
> abderraouf-adjal you are considering \0 as a char, but its a termination symbol for arduino. It should send your "P" in your callback message.

@ajaybnl Never mind about my first comment since I was confused with the index.

The issue is when all the buffer is used this might cause data loss in other variables. There is data, there is data length, it's all clear.

brianrho:
Not everybody's sending ASCII strings so asking for null termination is no good. In C/C++, if a length parameter is provided following a buffer, that's a strong indication that the buffer content may not be null-terminated. It's also not a good idea to encourage `payload[length] = 0` since you're writing to memory outside the specified length bounds, potentially overwriting useful data, especially if the library implementation changes at some point to hold multiple `PUBLISH` messages simultaneously in the buffer. If anything, I'd recommend the `payload` parameter be changed from `uint8_t*` to `const uint8_t*` to prevent this sort of thing.

PR #655, opened 2019-08-22
Using strnlen() instead of strlen()

abderraouf-adjal:
This will put a limit in case of inputs that are not null-terminated, or the array length is too long.
The modification replaced `strlen(buf)` with `strnlen(buf, MQTT_MAX_PACKET_SIZE)`.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -366,11 +366,11 @@ boolean PubSubClient::loop() {
 }
 
 boolean PubSubClient::publish(const char* topic, const char* payload) {
-    return publish(topic,(const uint8_t*)payload,strlen(payload),false);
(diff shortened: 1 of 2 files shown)


knolleary:
This doesn't really help for that very extreme edge case.

`MQTT_MAX_PACKET_SIZE` is the complete packet size, not just the payload.

Clipping a payload to that length won't help because it doesn't take into account the topic length or fixed header size - both of which are part of the packet.

abderraouf-adjal:
> Clipping a payload to that length won't help because it doesn't take into account the topic length

Of course.

In case of none null-terminated input this work fine, from perfectionism point of view we should calculate the exact limits which will be nice to _save few CPU cycles_;
but there is not much benefits from stop counting like 20 chars left in `strnlen()` input since the main goal here is to _protect against none null-terminated inputs_.

PR #661, opened 2019-08-31
Update README.md

cb1986ster:
Just a small hint in readme (it took me about 3h to get to this).


knolleary:
Thanks - but not adding hints like this here. There are endless numbers of systems you could use this library with - not going to start adding docs for them all.

PR #662, opened 2019-09-05
Removed the unused function yield

shivamag00:
Calling yield() on line 225 was creating problems while uploading the sketch LINKIT One

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -222,7 +222,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
 boolean PubSubClient::readByte(uint8_t * result) {
    uint32_t previousMillis = millis();
    while(!_client->available()) {
-     yield();
+     
      uint32_t currentMillis = millis();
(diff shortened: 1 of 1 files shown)


knolleary:
Why do you say it's unused? This line exists for a reason. You can't just remove it.

What problems does it cause?

shivamag00:
Today, I tried to upload a sketch, which used this library, to linkit one. On being uploaded, a warning was shown in the arduino IDE that in this file, there is undefined reference to the function yield(). And although the sketch was uploaded successfully, nothing was being done by LinkIt One. When I removed this function call from this cpp file and uploaded the sketch code again, no warning was shown. The sketch was uploaded successfully, and LinkIt One could subscribe and publish to various topics in my local Mosquitto Broker

shivamag00:
Sir, did you find the reason for calling the yield() function. I mean what it should do. Also, why has it not been defined anywhere?

knolleary:
https://www.arduino.cc/en/Reference/SchedulerYield

knolleary:
Closing as the `yield` function is there for a reason.

PR #665, opened 2019-09-24
Soft WDT reset fix on ESP8266

saloid:
I am getting `Soft WDT reset` while trying to use `connect(ID)`. This zero delay feeds watchdog timer while waiting `_client` answer.
ESP SDK config:
`SDK:2.2.1(cfd48f3)/Core:2.5.2=20502000/lwIP:STABLE-2_1_2_RELEASE/glue:1.1-7-g82abda3/BearSSL:a143020`

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -194,6 +194,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
                     _client->stop();
                     return false;
                 }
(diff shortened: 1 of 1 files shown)


saloid:
This (soft WDT reset) happens only if I am redefining yield()in my sketch. delay(0) fixes it anyway, but i don't know why it fails CI checking. delay() should be available since <Arduino.h> included.

PR #667, opened 2019-10-01
Some code cleanup

saloid:


src/PubSubClient.cpp
```diff
@@ -584,15 +581,15 @@ void PubSubClient::disconnect() {
...
-    const char* idp = string;
-    uint16_t i = 0;
-    pos += 2;
-    while (*idp) {
-        buf[pos++] = *idp++;
-        i++;
+    uint16_t copiedSymbolsNum = 0;
+    uint16_t lengthPosition = pos;
+    pos += 2;   //hold place for length (2 bytes)
+    while (*string == '\0') {
```
TD-er: This is not the same code.
The old code was checking for "not 0" and your code checks for "is 0"

src/PubSubClient.cpp
```diff
@@ -584,15 +581,15 @@ void PubSubClient::disconnect() {
...
+        buf[pos++] = *string++;
```
TD-er: Just a general remark.
Both in your change as in the original code is no check for the size of the array `buf`.

src/PubSubClient.cpp
saloid: Yes, thank you, fixed that.

src/PubSubClient.cpp
```diff
@@ -384,7 +383,7 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
             return false;
         }
         // Leave room in the buffer for header and variable length field
-        uint16_t length = MQTT_MAX_HEADER_SIZE;
+        uint16_t length;
         length = writeString(topic,buffer,length);
```
TD-er: The value for `length` is used as a parameter here, so if you don't initialize it, it will stop working very soon.

saloid: Done. Didn't notice that firstly it passed as argument. Now it should be more clear.

knolleary:
Does this actually fix any issues or is it just modifying code to a slightly different style? I'm wary of making changes for the sake of making changes.

TD-er:
> Does this actually fix any issues or is it just modifying code to a slightly different style? I'm wary of making changes for the sake of making changes.

Well at least this function is made a lot easier to read:
```c++
uint16_t PubSubClient::writeString(const char* string, uint8_t* buf, uint16_t pos) 
```

Not tested the code, but at first glance it looks to do the same.
My comment still applies by the way, unless it is clear it can never run out of the buffer length.

PR #685, opened 2019-11-29
fix/allow redefining constants (MQTT_MAX_PACKET_SIZE, ...)

d-a-v:
Redefining constants (like  `MQTT_MAX_TRANSFER_SIZE`) before including `PubSubClient.h` has no effect, because the library is precompiled with default values.
It is even dangerous because the compiled class used in sketch has mismatched valued with the externally (pre)compiled `PubSubClient.cpp` file.

This PR allows to force the compilation of the `.cpp` when including `PubSuClient.h`, using the same constants. There is no breaking change for the general case.

If a complex sketch would include `PubSubClient.h` from two different `.cpp` or `.ino` files, all source files except one would have to define `INCLUDE_CPP_FROM_PUBSUBCLIENT_H` to 0 before including.
Example:

`common.h`:
```cpp
#define MQTT_MAX_TRANSFER_SIZE 512 // application requirements
#include <PubSubClient.h>
```

`sketch-file-1.cpp`:
```cpp
#include "common.h"
```

`sketch-file-2.cpp`:
```cpp
#define INCLUDE_CPP_FROM_PUBSUBCLIENT_H 0
#include "common.h"
```

This is a simple working hack.
The correct way for doing that would be to transform these defines/constants as class members, with the matching getters and setters. This would (breaking) change the API and make the code a little bigger. If the author wishes to, I can update my PR in that way.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -425,7 +425,7 @@ boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsig
     unsigned int i;
     uint8_t header;
     unsigned int len;
-    int expectedLength;
+    unsigned int expectedLength;
 
     if (!connected()) {
         return false;
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -5,7 +5,9 @@
   http://knolleary.net
 */
 
-#include "PubSubClient.h"
+//#include "PubSubClient.h"
+#if INCLUDE_CPP_FROM_PUBSUBCLIENT_H
+
 #include "Arduino.h"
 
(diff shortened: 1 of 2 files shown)


knolleary:
Thanks for suggesting this approach - but as you say, its a bit too much of a hack and not one I'm keen on adopting.

I've voiced my reasons lots of times for keeping the constants statically defined in the header, rather than allowing them to be dynamically set. I think some of the reasons for that position are less relevant these days given the typical hardware is much more capable than when this library was first written - so I may look at it again at some point. But this library doesn't get a huge amount of my time and focus as I have some other more demanding OSS projects to keep working.

In the interim, I'm considering upping the default packet size to `512`.

PR #715, opened 2020-02-29
Added error reporting

pmkevans:
Adding more information about what's going on to facilitate fault finding in release builds.
Harmonised coding style.


knolleary:
Hi, can you explain more about what this PR does? In what way does it provide more information? Can you provide an example?

I see you've also chosen to unilaterally reformat all of the code without any explanation. That makes it hard to see what functional changes you are proposing. It is also not good etiquette to just reformat the whole code base without first discussing it. Why should I adopt your preferred style over what was already there?

pmkevans:
The main thing is extending the class interface to include PubSubClient::getLastStatus(), and then populating the underlying buffer with information about what's going on.  ie: More result code checking and explanation.
I acknowledge the code reformatting should have been  a separate request.  The existing code is a bit inconsistent in naming and calling convention; I picked one style that was already present and adopted that throughout for readability. 
I wasn't originally planning to contribute back to your code base but someone suggested I was being selfish in not doing so.  No drama if you don't want it.  Thanks for your work so far.

knolleary:
In this instance, reformatting the code has broken all of the unit tests.

I can see merit in having the extra debug, but I can't merge with the reformatting.

knolleary:
How much larger is a sketch compiled with these changes added? Keeping the library as small as possible is an important requirement.

pmkevans:
I'm not certain what additional files you're referring to.  It's additionally pulling in stdio.h but that's all(?).
If you'd like to nominate a code standard I'm happy to work with it.  If reformatting for consistency is off the table we can call it quits now.
Beyond that, I can see the Arduino IDE isn't going to cut it for contributing so there's a bigger exercise there for me which I won't get to for a while.

knolleary:
You've added lots of files under a `.vs` directory, such as `.vs/ProjectSettings.json` and other binary files.

If you want to contribute a functional change, please do that without forcing a style change on the project.

PR #727, opened 2020-04-11
Added check to prevent subscribe/unsubscribe to empty topics

sn0kerbuzz:
When subscribing (unsubscribing) to an empty topic the client disconnects with state `MQTT_CONNECTION_LOST`.


knolleary:
Thanks!

PR #733, opened 2020-05-01
Use strlen_P instead of strlen as you are reading from PROGMEM.

jasiek:
Tested with Arduino Uno & Ethernet shield.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,14 @@
 sudo: false
-language: cpp
-compiler:
-  - g++
-script: cd tests && make && make test
+language: python
+python:
+  - "2.7"
+cache:
+  directories:
+    - "~/.platformio"
+install:
+  - pip install -U platformio
+  - platformio update
+script:
+  - platformio run
 os:
   - linux
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,22 @@
-sudo: false
-language: python
-python:
-  - "2.7"
-cache:
-  directories:
-    - "~/.platformio"
-install:
-  - pip install -U platformio
-  - platformio update
-script:
-  - platformio run
-os:
-  - linux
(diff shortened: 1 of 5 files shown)


knolleary:
Thanks for this.

Unfortunately I cannot merge it as-is because it modifies more than just the strlen_P fix.

Can you back out those other changes, or at least explain what they are doing and why you've made them - including why you've stopped it from running the unit tests?

jasiek:
Sure.

I've adjusted the Travis build script to build the project using platformio (platformio.org). This will build it against a particular board/architecture. The aim of this is to check whether this actually builds on an Arduino. The progmem.ino example is a part of that - a program which does nothing but references functions used. Now that I think about it - that test suite compiles just fine, so that ought to be brought back.

knolleary:
If you want me to consider moving this library to a different CI system, then it needs to be proposed and discussed in its own right. Please don't unilaterally change it as a side effect of an unrelated bug fix.

The minimal example you've added is of no use to an end user. It is not an example they can use.

jasiek:
I think some complexity lies ahead - your unit tests will run on my desktop, no problem. Those unit tests won't be able to compile if we include a reference to `<avr/pgmspace.h>` and the `strlen_P` function.

So the way to solve this would be to compile it against the actual libraries it would be compiled against when running on the board. PlatformIO would solve this problem quite nicely, IMO.

The downside of that is that unit tests would have to run on a simulated device, and if this library has the ambition to work with many platforms, it'd be a good idea to test it against all of them. PlatformIO has the ability to run unit tests on the real device itself, even over the internet.

One way to avoid all that complexity is to remove functionality related to PROGMEM. It's not implemented correctly, no one seems to care about the bug, so that'd mean that no one is *really* using it.

What do you think?

PR #746, opened 2020-06-10
Added the possibility to define the MQTT topis as F() string

t-dan:
Could you please consider this patch, allowing to pass an `F()`-defined strings as a MQTT message topic.

Thank you.
Dan T.


knolleary:
How does this compare with the existing `publish_P` function for publishing a payload from PROGMEM?

I appreciate it doesn't take the topic from PROGMEM, but it feels a little odd to have pulled `publish_P` out as a separate function, and then to overload `publish` with this change.

t-dan:
Regarding the `publish_P()`, it seems to me that it accepts PROGMEM strings as a payload, not as the topic. And anyway, if I understand it well, there must be `__FlashStringHelper *` input argument to use:
`client.publish(F("/topic/"), "a payload")`

I have tried the code on my Arduino, and it seems to work. The use of `publish_P()` did not work for me ...

.

knolleary:
Yes, I understand the existing publish_P function doesn't support topics from PROGMEM. But it has set the principle of separating out the functions that use PROGMEM from those that don't.

From an API design, adding `topic` PROGMEM support to the `publish` functions that don't currently do *anything* with PROGMEM, but not touching the `publish_P` function makes things a bit awkward.

There's another PR that addresses some of the functional issues with the `publish_P` function - it just pulls in some baggage that I'm reluctant to merge.

I have no issue with adding support for topics coming from PROGMEM - I just want to make sure the overall API feels consistent.

t-dan:
Just a question to be clear that I understand it.
Is the `publish_P()` method supposed to work with PROGMEM string in a topic or a payload?

My use case is, that the topics are fixed (thus the use of `F()`) and the paylod differs call to call (and needs to use classic memory strings then).

If this matches the design of `publish_P()`, my proposed implementation could be "optimized" a bit to use `publish_P()` (after the pending patches are applied). And the code can be deduplicated a bit as well. I may have a look at it, if you want to.

t-dan:
And a side note:
as far as I understand it, the functions accepting PROGMEM strings accept `const char *` and are suffixed with _P in Arduino. Functions accepting `F()` strings are not suffixed, but accept `const __FlashStringHelper*`.

I thing that it matches your interface design.

t-dan:
Hallo,
I have made little code cleaning (based on the previous patch), mostly the replacement of string copy loops with `strncpy()`/`memcpy()`. The idea is that using these functions de-duplicate the code and there is also an assumption that their implementation may be a bit faster (in the worst case, they are implemented as loops anyway). Plus there is some removal of repeated `strnlen()` calls.

All is prepared, tests passed. It has also been intensively tested on my Arduino, sending MQTT messages there and back for a half of a day.

But I think that pushing it to my branch will add it to this MR. So, should I post it anyway, or wait a bit longer?

Thank you,
DT

Issue #300, opened 2017-07-07
setCallback in custom library error


msnederland:
Hi,

I'm writing a custom library sourcing PubSubClient. Everything went well expect for integrating the callback function in my library. 

I get the following errors from the Arduino IDE:

>Error: no matching function for call to 'PubSubClient::setCallback("<"unresolved overloaded function type">")' client -> setCallback(callback);

In my xxx.h file I declared the following:

`class Iotdevice {`
...
    `void callback(char* topic, byte* payload, unsigned int length);`
...
`}`

In my xxx.cpp I defined the following:

`void Iotdevice::callback(char* topic, byte* payload, unsigned int length) {...)`

I have tried a lot of variations, but I can't seem to solve it. Can anyone help me out?

DeltaCore:
You can't use a class function for the callback. Define a global function and then call the class function from there.

Pimmetje:
Correct but wouldn't it be nice to if u could overload the method in a subclass.

DeltaCore:
@Pimmetje That would be nice to. Haven't tried that one.

Pimmetje:
I tried it. I made some small changes to the library and it worked for the simple tests i did. But it requires your to inherit from PubSubClient in order to overload the method of course. 
I was working on this to make a base class that would handle wifi mqtt connecting and reconnecting without reprogramming it for every project. I also added a http update server just for fun but now i am getting off topic.

sduriez:
Hi,
What modifications had you made to PubSubClient to use the callback in another Class?
Thanks

Simon

Pimmetje:
Just a quick and dirty writeup (untested). If someone wants a proper diff let me know but it ill take me a little more time :D

Relevant part of the loop code: note the this->_callback lines:
```
                    if ((buffer[0]&0x06) == MQTTQOS1) {
                        msgId = (buffer[llen+3+tl]<<8)+buffer[llen+3+tl+1];
                        payload = buffer+llen+3+tl+2;
                        this->_callback(topic,payload,len-llen-3-tl-2);

                        buffer[0] = MQTTPUBACK;
                        buffer[1] = 2;
                        buffer[2] = (msgId >> 8);
                        buffer[3] = (msgId & 0xFF);
                        _client->write(buffer,4);
                        lastOutActivity = t;

                    } else {
                        payload = buffer+llen+3+tl;
                        this->_callback(topic,payload,len-llen-3-tl);
                    }
```

Added methods (h file)
```
private:
[...]

PR #760, opened 2020-07-28
Allow inheritance of PubSubClient into a custom class

xluthi:
A specific callback signature is required to allow full inheritance of 
the PubSubClient class. Inheritance allows a better code separation by 
putting all the MQTT logic within one custom class (including the 
callback method!).

The code is based on a proposal in issue knolleary/pubsubclient#300, and could close this issue.

Any comment is welcome :-)

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -393,29 +393,27 @@ boolean PubSubClient::loop() {
                 lastInActivity = t;
                 uint8_t type = this->buffer[0]&0xF0;
                 if (type == MQTTPUBLISH) {
-                    if (callback) {
-                        uint16_t tl = (this->buffer[llen+1]<<8)+this->buffer[llen+2]; /* topic length in bytes */
-                        memmove(this->buffer+llen+2,this->buffer+llen+3,tl); /* move topic inside buffer 1 byte to front */
-                        this->buffer[llen+2+tl] = 0; /* end the topic as a 'C' string with \x00 */
-                        char *topic = (char*) this->buffer+llen+2;
-                        // msgId only present for QOS>0
-                        if ((this->buffer[0]&0x06) == MQTTQOS1) {
-                            msgId = (this->buffer[llen+3+tl]<<8)+this->buffer[llen+3+tl+1];
-                            payload = this->buffer+llen+3+tl+2;
-                            callback(topic,payload,len-llen-3-tl-2);
+                      uint16_t tl = (this->buffer[llen+1]<<8)+this->buffer[llen+2]; /* topic length in bytes */
+                      memmove(this->buffer+llen+2,this->buffer+llen+3,tl); /* move topic inside buffer 1 byte to front */
+                      this->buffer[llen+2+tl] = 0; /* end the topic as a 'C' string with \x00 */
+                      char *topic = (char*) this->buffer+llen+2;
+                      // msgId only present for QOS>0
+                      if ((this->buffer[0]&0x06) == MQTTQOS1) {
+                          msgId = (this->buffer[llen+3+tl]<<8)+this->buffer[llen+3+tl+1];
+                          payload = this->buffer+llen+3+tl+2;
+                          _callback(topic,payload,len-llen-3-tl-2);
 
-                            this->buffer[0] = MQTTPUBACK;
-                            this->buffer[1] = 2;
-                            this->buffer[2] = (msgId >> 8);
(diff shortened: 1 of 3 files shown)


xluthi:
By the way, the proposed change is backward compatible, as it rely on the "legacy" callback method in case the new virtual one is not defined in any child class.

eku:
https://en.wikipedia.org/wiki/Composition_over_inheritance

xluthi:
> https://en.wikipedia.org/wiki/Composition_over_inheritance

Thanks @eku for the comment.  Composition over inheritance is a design pattern that could have some benefits in some cases.  However, due to the specific signature of the original callback function, it must always be a global function.

This pull request allows to have the callback function defined in a specific namespace (i.e. in a class inherited from PubSubClient), hence allowing a proper code separation (at least from my point of view ;-) ).

eku:
I do not quite understand what is meant by separation.

The class can implement the callback signature of the ```PubSubClient``` and offer another callback interface to the outside. You don't have to inherit for this. It is sufficient if the class encapsulates the ```PubSubClient``` as a member without passing its API directly.

src/PubSubClient.cpp
```diff
@@ -393,29 +393,27 @@ boolean PubSubClient::loop() {
...
-                            this->buffer[3] = (msgId & 0xFF);
-                            _client->write(this->buffer,4);
-                            lastOutActivity = t;
-
-                        } else {
-                            payload = this->buffer+llen+3+tl;
-                            callback(topic,payload,len-llen-3-tl);
-                        }
-                    }
+                      uint16_t tl = (this->buffer[llen+1]<<8)+this->buffer[llen+2]; /* topic length in bytes */
```
eku, line 396: The following lines are always executed, even if no callback is assigned. Is this intentional?

xluthi: Indeed, the check on the "callback" assignment is delegated within the new _callback function, but functionally, the behaviour is the same.

xluthi:
> I do not quite understand what is meant by separation.

This means, for example, that you can write all the code related to the MQTT interface within a dedicated class, in a dedicated file, and only call the main methods within your sketch.  All the MQTT code is inside one specific object.

> The class can implement the callback signature of the `PubSubClient` and offer another callback interface to the outside. You don't have to inherit for this. It is sufficient if the class encapsulates the `PubSubClient` as a member without passing its API directly.

In this case, the compositing class will still rely on a global function to implement the callback of the PubSubClient (callback cannot be an object method).  Furthermore, we should define in this class all the methods of PubSubClient that I need to be exposed: this is one of the main drawback of composition.

As a conclusion, I will also send you a Wikipedia Link: [TIMTOWTDI](https://en.wikipedia.org/wiki/TIMTOWTDI) :-)

eku:
> In this case, the compositing class will still rely on a global function to implement the callback of the PubSubClient (callback cannot be an object method).

Is this a limitation of the Arduino framework? 

As far as I know, the callback method would only have to be declared ```static``` to be passed as callback to the ```PubSubClient```.

xluthi:
> > In this case, the compositing class will still rely on a global function to implement the callback of the PubSubClient (callback cannot be an object method).
> 
> Is this a limitation of the Arduino framework?
> 
> As far as I know, the callback method would only have to be declared `static` to be passed as callback to the `PubSubClient`.

There is no specific limitation of the Arduino framework, but you mentioned it correctly: the callback method must be a static one. It cannot be a non static class method.  It means it will not have access to any non-static class variables, and that in one sketch, any instance in that class would share the same callback function.

With my proposal, I could have easily two instances of the inherited class with specific callback method for each.

In this case, I see only benefits to allow a proper inheritance of the PubSubClient class, hence my pull request proposal.

PS: see https://github.com/xluthi/pulse_counter_esp8266 for a "real" example, with all the MQTT logic within the [MqttBackend.h](https://github.com/xluthi/pulse_counter_esp8266/blob/e4af92d452c8e80667952a003c70696682dacc6e/MqttBackend.h) file (and .cpp).

eku:
Thanks for the explanation. Under the circumstances described, PR makes sense.

src/PubSubClient.cpp
eku: Yes, but only as long as a callback is set. You can discuss now whether the client without callback is useful, but that was not my point.

marcelo-lara:
It's an old PR are there any plans to merge?
thanks!

arminsalcin:
When this will be merged ?

PR #761, opened 2020-07-29
Add loop after initial connect in reconnect example

nelsonblaha:


eku:
Would you please describe why this change is necessary?

PR #774, opened 2020-09-10
Expose Outstanding Actions

cBiscuitSurprise:
In a recent project, I ran into an issue where my Arduino would routinely lock up without warning and at random times. I was able to finally trace it down to MQTT responses coming during the middle of a publish. This was causing SSLClient to lock up and thus my Arduino to lock up.

The two points where my particular application ran into this issue were with keep alive pings and subscribes. To get around my immediate issue, I just exposed `pingOutstanding` (moved from `private:` to `public:`) and created `subscribeOutstanding`. I then just wait until those are both `false` before trying to publish.

This should probably be generalized a little more with accessor functions rather than using parameters directly, but I just wanted to get something working. I'm willing to work on making this more "productionized" and generic (to include all the ACK callbacks), but wanted to get some guidance from the team about the best path forward.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -427,9 +427,6 @@ boolean PubSubClient::loop() {
                     subscribeOutstanding = false;
                 } else if (type == MQTTPUBACK) {
                     publishOutstanding = false;
-                } else {
-                    Serial.print("Unhandled response: ");
-                    Serial.println(type);
                 }
             } else if (!connected()) {
                 // readPacket has closed the connection
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -422,6 +422,8 @@ boolean PubSubClient::loop() {
                     _client->write(this->buffer,2);
                 } else if (type == MQTTPINGRESP) {
                     pingOutstanding = false;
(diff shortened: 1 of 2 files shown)


cBiscuitSurprise:
Then, in my application, I have this for every `loop` call:

```c++
void loop_client() {
  client.loop();

  unsigned long pingWaitStart = millis();
  while (client.pingOutstanding) {
    if (millis() - pingWaitStart > 1000) {
      Serial.println("[mqtt] Warning, timed out waiting for PING-RESPONSE from broker.");
      break;
    }
    client.loop();
  }
}
```

... and similar for each subscribe:
```c++
void subscribe_one(const char *topic) {
  client.subscribe(topic);

  unsigned long subWaitStart = millis();
  while (client.subscribeOutstanding) {
    if (millis() - subWaitStart > 1000) {
      Serial.println("[mqtt] Warning, timed out waiting for SUB-ACKNOWLEDGEMENT from broker.");
      break;
    }
    client.loop();
  }

  Serial.print("[mqtt] subscribed: ");
  Serial.println(topic);
}
```

Obviously, people that don't like blocking could do it differently, but this works for me for now (I have a patient watchdog ... 😄).

PR #780, opened 2020-09-30
Add ESP8285 to supported devices

Flole998:
Add ESP8285, which is basically an ESP8266 with integrated flash memory

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ boards and shields, including:
  - TI CC3000 WiFi - [library](https://github.com/sparkfun/SFE_CC3000_Library)
  - Intel Galileo/Edison
  - ESP8266
+ - ESP8285
  - ESP32
 
 The library cannot currently be used with hardware based on the ENC28J60 chip –


rvt:
Perhaps it should say ESP architecture or something in the line like that?

Flole998:
That's not a good idea, once new devices appear they might need to be excluded as compatibility isn't tested and verified yet. And as you can see on this PR even such a simple change takes a long time, so wrong information would be in the Readme file for quite some time.

knolleary:
I have no way of verifying any claim as to what this client works on or not, beyond what I have personally tested. If I claim "support" for a particular device, there is an implicit assumption I'll help debug any issue that is hit on that platform. That is unlikely to happen.

I am more inclined to update the README to say something more generic, rather than list out each device that exists.

I would pretty much expect it to just work with any device supported by the Arduino IDE - any hardware specific parts are in the underlying network client library that is nothing to do with this library.

PR #783, opened 2020-10-04
Fix bug in publish_P which will always treat the payload length as 0 when used with PROGMEM

tomer-w:
publish_P used to call strlen to calculate the payload length. This will not work on real PROGMEM memory.
The solution is to call strlen_P which will calculate the string length correctly.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -472,7 +472,7 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
 }
 
 boolean PubSubClient::publish_P(const char* topic, const char* payload, boolean retained) {
-    return publish_P(topic, (const uint8_t*)payload, payload ? strnlen(payload, this->bufferSize) : 0, retained);
+    return publish_P(topic, (const uint8_t*)payload, payload ? strnlen_P(payload, MQTT_MAX_POSSIBLE_PACKET_SIZE) : 0, retained);
 }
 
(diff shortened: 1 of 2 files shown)


tomer-w:
Any maintainer who can approve this one is available to check this out?

knolleary:
Hi @tomer-w  - I try to check this out in the next few days. I don't have a lot of spare time at the moment.

hmueller01:
@knolleary can you please merge this. It took me two nights to debug this and find the same solution like @tomer-w . Just wanted to open a new PR and found this. Oh man. Please please merge to avoid more people spending a lot of time. THANKS!

hmueller01:
But wait, why did you replace `this->bufferSize` by `MQTT_MAX_POSSIBLE_PACKET_SIZE`?
I think this is not good, as bufferSize is set by `boolean PubSubClient::setBufferSize(uint16_t size)` and reserves the buffer in memory ...

tomer-w:
The buffet which is referred by setBufferSize is used to receive MQTT packets. The publish_P function does not use it and therefore not relevant in this case.

hmueller01:
Ok, you are right. The `this->buffer[pos++]` is only used for the header and the length. The payload is written char by char without use of the buffer.
`rc += _client->write((char)pgm_read_byte_near(payload + i));`
I think it looks good.

tomer-w:
Yep :) I really hope this will get merged after 2 years.

Issue #795, opened 2020-11-21
KeepaliveTimeout error with different brokers


hutje:
When I connect to the MQTT server it works, but after 10 - 30 seconds the broker gives a keepalive timeout and reconnects rightaway. The loop is continuously executed. 
I tried different brokers but still the same problem. Mosquitto and Aedes (Node-Red) both are set to 60 seconds by default.
So the broker expects every 60 seconds a PINGREQ.
The esp8266 are default set to 15 seconds in the PubSubClient Library.
So every 15 seconds it sends a PINGREQ. The code is without any delay so client.loop() is constantly being updated.
Why do I still get all these disconnections?

```
void loop(){
 reconnectWifi();
 client.loop();
}
```

```
void reconnectWifi() { //(re)connect to WiFi and MQTT Server.
  if (WiFi.status() != WL_CONNECTED && currentMillis - wifiLostTimer >= 500UL ) {
    wifiLostTimer = currentMillis;
    delay(0); //to keep wifi connection alive same als yield()
    Serial.println("Wifi connecting");
    wifiLost = true;
    mqttLost = true;
    reconnectTried++;
    if (reconnectTried == 60 && digitalRead(relaySwitch) == relayOffOutput) {
      ESP.restart();
    }
  }

  else if (WiFi.status() == WL_CONNECTED && wifiLost == true) {
    reconnectTried = 0;
    wifiLost = false;
    Serial.println("Wifi Connected!!!");
  }
  // Loop until we're reconnected
  else if (client.connected() == false && currentMillis - timeOut >= 10000UL && WiFi.status() == WL_CONNECTED) {
    mqttLost = true;
    delay(0); //to keep wifi-connection alive
    timeOut = currentMillis;
    Serial.println("Attempting MQTT connection...");
    // Attempt to connect
    Serial.println(deviceID.c_str());
    client.connect(deviceID.c_str());
  }
  else if (client.connected() == true && wifiLost == false && mqttLost == true) {
    mqttLost = false;
    Serial.println("MQTT Connected!!!");
    String rxTopic = rxTopicFirst + deviceID;
    client.subscribe(rxTopic.c_str());
    Serial.println("Subscribed to: " + rxTopic);
  }
}
```

hutje:
Found that the problem is caused by bad wifi-connection. This way the package takes too long to send.
I checked the RSSI and the problems occur from -77 and higher.

hutje:
[...]

PR #802, opened 2020-12-23
Fixes timeout issues like stated in issue #795

roberthend:
Fixes the timeout issues, tested on own project and no timeouts for over two weeks.


roberthend:
> Why does this fix the problem? Is it a quick-fix or is it a legitimate fix that does not have any consequences?

When a connection is made the pingOutstanding flag is used to keep track if the client has send and ping and is waiting for a reply (in the loop function). When connection is lost the clients reconnects but fails to set pingOutstanding to false. This causes a state where a connection is made to the broker and the client keeps waiting for a pingresponse without sending a pingrequest leading to multiple reconnects after the keepalive period is hit. This sets the pingOutstanding flag to false when the client disconnects so on reconnect a pingrequest is actually send and that fixes the keepalive timeout loop.

PR #817, opened 2021-01-22
Added QOS2 Support (Asynchronous)

shakil-anwar:
I have added Qos2 full support. A buffer of maximum 10 is set by default.  and a function(PubSubClient::qos2Emptry(()) is added to check wheather the buffer is full or not. Buffer will be full if it fails to send all 10 data String. So One have to keep those unconfirmed data for qos2. You can change max qos2 buffer size by changing MQTT_QOS2_MAX_BUFFER value. Working great.


knolleary:
Hi - you should maybe have asked before spending your time on this. There are a variety of reasons why this library doesn't support QoS2. For example, a lack of any proper persistence of messages means most of the QoS2 semantics cannot be provided.

You'll need to provide much more information about the implementation, what new APIs you've added and how they are meant to be used before I'll look at this PR.

PR #835, opened 2021-02-27
Resolve issue #832 (Potential memory corruption)

arjenhiemstra:
The memmove command is executed without any sanity checking on used values causing random reboots when an MQTT package gets processed with an impossible topic length.
With a high rate test flow of MQTT messages this would sometimes corrupt memory after a few seconds.

Extra sanity checks and calculation for payload_offset and message length have been implemented to better handle messages and prevent this memory corruption.

Tested with a flow of about 50.000 messages per hour for multiple days. Went from hundreds of errors per hour to a few per day, the last errors mainly because the MQTT server couldn't keep up with the message flow. Never crashed again.
A big thanks to @TD-er for help with further refactoring the code.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -367,7 +367,82 @@ uint32_t PubSubClient::readPacket(uint8_t* lengthLength) {
     return len;
 }
 
+boolean PubSubClient::loop_read() {
+    if (_client == nullptr) {
+        return false;
+    }
+    if (!_client->available()) {
+        return false;
+    }
+    uint8_t llen;
+    uint16_t len = readPacket(&llen);
+    if (len == 0) {
+        return false;
+    }
+    unsigned long t = millis();
+    lastInActivity = t;
+    uint8_t type = buffer[0]&0xF0;
+
+    switch(type) {
+        case MQTTPUBLISH: 
+        {
+            if (callback) {
+                const boolean msgId_present = (buffer[0]&0x06) == MQTTQOS1;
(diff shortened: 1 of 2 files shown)


src/PubSubClient.cpp
```diff
@@ -368,7 +368,7 @@ uint32_t PubSubClient::readPacket(uint8_t* lengthLength) {
 }
 
 boolean PubSubClient::loop_read() {
-    if (_client == nullptr) {
+    if (_client == NULL) {
```
TD-er, line 371: Why do this change?
`NULL` is not a pointer type, so the compiler cannot help you here to check for wrong usage.
`nullptr` is a pointer type, so it is an improvement over the old `NULL`

arjenhiemstra: Just because otherwise the travis ci build behind the repo doesn't run

TD-er: Hmm that's strange.
What kind of compiler or compiler settings is used?

arjenhiemstra:
if you click on details behind "continuous-integration/travis-ci/pr — The Travis CI build failed"
you see all the build system info

src/PubSubClient.cpp
arjenhiemstra: if you click on 'details' behind "continuous-integration/travis-ci/pr — The Travis CI build failed"
you see all the build system info

technyon:
Any update on this PR? It it going to get merged?

PR #840, opened 2021-03-16
Initialize callback to NULL to avoid exception on calling uninitialized callback

Bascy:
Not initializing the callback pointer will cause null pointer exceptions if it is not set manually. If not initialized it will (probably) not be 0 and therefore i.e. PubSubClient.cpp, line 405 will try to call a random address

Some constructors actually set the callback to NULL but most are not.
To make sure that callback will always be set to NULL unless overwritten in a specific constructor or by calling setCallBack() it should be initialized on declaration

### DIFF
--- a/src/PubSubClient.h
+++ b/src/PubSubClient.h
@@ -96,7 +96,7 @@ private:
    unsigned long lastOutActivity;
    unsigned long lastInActivity;
    bool pingOutstanding;
-   MQTT_CALLBACK_SIGNATURE;
+   MQTT_CALLBACK_SIGNATURE = nullptr;
    uint32_t readPacket(uint8_t*);
    boolean readByte(uint8_t * result);
    boolean readByte(uint8_t * result, uint16_t * index);
--- a/src/PubSubClient.h
+++ b/src/PubSubClient.h
@@ -96,7 +96,7 @@ private:
    unsigned long lastOutActivity;
    unsigned long lastInActivity;
    bool pingOutstanding;
-   MQTT_CALLBACK_SIGNATURE = nullptr;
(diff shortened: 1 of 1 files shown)


knolleary:
Please add a description to explain what this change is and why it is necessary. Sure it looks simple enough, but *why* is that needed?

Bascy:
Not initializing the callback pointer will cause null pointer exceptions if it is not set manually, but on closer examination i see that constructors initialize the pointer ..
will have to investigate why its causing NPE in my project

Bascy:
Updated original comment to explain

hmueller01:
I would rather use `setCallback(NULL);` in all PubSubClient constructors that do not contain the `MQTT_CALLBACK_SIGNATURE` parameter. Currently only `PubSubClient()`does it. 

What do you think about changing all these repetitions of 
```
PubSubClient::PubSubClient(Client& client) {
    this->_state = MQTT_DISCONNECTED;
    setClient(client);
    this->stream = NULL;
    this->bufferSize = 0;
    setBufferSize(MQTT_MAX_PACKET_SIZE);
    setKeepAlive(MQTT_KEEPALIVE);
    setSocketTimeout(MQTT_SOCKET_TIMEOUT);
}
```
to
```
PubSubClient::PubSubClient(Client& client) {
    PubSubClient(); // default initialize all members
    setClient(client);
}
```
and so on ...

Bascy:
I dont think this repo is maintained anymore ... last changes are from 4 years ago

hmueller01:
There is a similar approach here #746. @knolleary is willing to do some minor changes #1045, not sure if this will make it.

PR #841, opened 2021-03-18
size is now 32 bit not 16

mirecta:
I send to mqtt  jpeg image around 100kB  (ussing message pack its not important) but due limitation size to 16 bit pubsub client cant write regular size in mqtt header


mirecta:
i continue with my fork which is fully streamed i can send more than 65K and receive too , i ve made OTA through MQTT

PR #901, opened 2021-10-20
add delay that that prevent watchdog to crash

clecle253:


clecle253:
Hello, I'm using your library on a project which uses an ESP32 with esp-idf and arduino.

During the dev I encountered an issue with the `connect()` function:
The `while (!_client->available())` loop can sometimes loop very quickly. While that's no issue on some MCUs, the esp32 has a task watchdog system: a thread cannot take 100% of the cpu time, and needs to yield once in a while, or the code will crash.

This PR adds a 1ms delay between retries to give time to other threads to do their work, thus fixing the problem.

Flole998:
Most likely what you really want is to `yield()`.

PR #907, opened 2021-11-26
Added will message length to connect method.

iojan:
First, thanks for this great library!
Current connect method only allows null terminated strings as _willMessage_ payload. 
I needed to send binary data, so I added a _willMessageLenth_ argument. 
If _willMessageLenth_ is zero, willMessage will be copied as a null terminated string. If bigger than zero, _willMessageLenth_ bytes will be copied form _willMessage_.
To keep it simple, buffer length check and copy were implemented inside the connect method. I can create a new check macro and copy function if you prefer.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -179,6 +179,10 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
 }
 
 boolean PubSubClient::connect(const char *id, const char *user, const char *pass, const char* willTopic, uint8_t willQos, boolean willRetain, const char* willMessage, boolean cleanSession) {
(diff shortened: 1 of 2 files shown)


Illirgway:
Wouldn't it be better to use memcopy / memmove instead of direct byte-to-byte copy loop?

```
                // const uint8_t* willMsg, uint16_t willMsgSz
                // check willMsg len
                // TODO check overflow
                if (length + 2 + willMsgSz > this->bufferSize) {
                  _client->stop();
                  return false;
                }

                // store msg size prefix
                this->buffer[length+0] = (willMsgSz >> 8);
                this->buffer[length+1] = (willMsgSz & 0xFF);
                length += 2; // count size prefix

                memcpy(&this->buffer[length], willMsg, willMsgSz); // or memmove

                length += willMsgSz;
```

PR #1001, opened 2023-03-01
Fix: Vulnerability in Comparison of different type sizes

FeedehC:
Little change in sizes of for loop iterator "i", from uint16_t to uint32_t, to avoid possible infinite loop, which depends on the value of "plength".
More information in issue: https://github.com/knolleary/pubsubclient/issues/1000


Flole998:
Well, now `length` could potentially overflow so the buffer is incorrectly filled. Also you're still mixing datatypes, an `unsigned int` and an `uint32_t` could potentially have different sizes, so I suggest to use `unsigned int` for the length everywhere.

PR #1005, opened 2023-03-17
Minor unsigned int type correction (avoid warning)

smartynov:
Changed `expectedLength` to `unsigned int` to avoid confusion and compiler warning


hmueller01:
This is fixed in
Maybe you give it a try.

PR #1020, opened 2023-07-29
fix: use of wrong strnlen variant on PROGMEM payload strings

### LINKED ISSUE #747: esp8266 crashes when using payload in PROGMEM (v2.8)
SInce the latest update (v2.8) 
`publish_P(const char* topic, const char* payload, boolean retained)`
crashes when using it with payload located in PROGMEM.

Example: Using of 
`client.publish_P("outTopic_P", PSTR("test"), true);`
was possible in v2.7, but no longer in v2.8. (tested with [mqtt_esp8266](https://github.com/knolleary/pubsubclient/blob/master/examples/mqtt_esp8266/mqtt_esp8266.ino))

Seems that the problem is in line 475
`return publish_P(topic, (const uint8_t*)payload, payload ? strnlen(payload, this->bufferSize) : 0, retained);`
Maybe better use strnlen_P. 

Thank you.

muggenhor:
strnlen_P is needed to avoid the wrong method of reading to cause bad things to happen (CPU exceptions on ESP8266).

Fixes: #747

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -472,7 +472,7 @@ boolean PubSubClient::publish(const char* topic, const uint8_t* payload, unsigne
 }
 
 boolean PubSubClient::publish_P(const char* topic, const char* payload, boolean retained) {
-    return publish_P(topic, (const uint8_t*)payload, payload ? strnlen(payload, this->bufferSize) : 0, retained);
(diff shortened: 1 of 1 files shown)


hmueller01:
This is fixed in
Maybe you give it a try.

PR #1022, opened 2023-08-02
Quick fix for Arduino R4

slviajero:
fixed the problem with missing strnlen on Arduino R4 Wifi by adding strnlen into PubSubClient.h and PubSubClient.cpp only for this platform. This is ugly but works.


slviajero:
In the meantime, a way to fix this in Arduino platform.txt has been found. Probably not needed any more.

PR #1031, opened 2023-10-07
change int to unsigned int to avoid compiler warmings in PubSubClient.cpp

wilmsn:
Change variable type (int to unsigned int) to avoid compiler warnings.


hmueller01:
This is solved in https://github.com/hmueller01/pubsubclient3

Issue #1033, opened 2023-11-12
Issue while setting keep alive zero.


kabirwbt:
On my esp8266, If I set the keep alive parameter to zero and connect to a broker, despite being connected to broker successfully, connected() returns false and my reconnect loop triggers automatically. I compiled the example provided in the following link : https://github.com/knolleary/pubsubclient/blob/master/examples/mqtt_esp8266/mqtt_esp8266.ino.

I just added the following line to void setup(), at the absolute bottom: client.setKeepAlive(0);

`
Attempting MQTT connection...connected
Attempting MQTT connection...connected
`
This message is printed in a rapid manner, infinitely. 

Am I doing something wrong? Should connected() not return true in my case?

kabirwbt:
My guess is that some other block of code, maybe the ping mechanism is disconnecting the client and causing the connected() to return false. 
EDIT:
after some digging:
https://github.com/knolleary/pubsubclient/blob/2d228f2f862a95846c65a8518c79f48dfc8f188c/src/PubSubClient.cpp#L376 is causing the connection to close. Shouldn't the pinging mechanism be disabled via setting the keepalive parameter to 0 ?

ChuckMash:
Yep, looks like keepalive 0 is not correctly respected.


Try changing the following line 

https://github.com/knolleary/pubsubclient/blob/2d228f2f862a95846c65a8518c79f48dfc8f188c/src/PubSubClient.cpp#L373
to
```
if (((t - lastInActivity > this->keepAlive*1000UL) || (t - lastOutActivity > this->keepAlive*1000UL)) && keepAlive != 0) {
```

I've not done any long term testing, but I'm having no issues with this change so far.

PR #1038, opened 2023-12-10
bugfix for keep alive zero

ChuckMash:
Setting keep alive to 0, or infinite keep alive, is not currently handled correctly.
Currently the client will immediately disconnect when keep alive is set to 0, this fix adds a check for this state before attempting PINGREQ.

Solves #1033

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -370,7 +370,7 @@ uint32_t PubSubClient::readPacket(uint8_t* lengthLength) {
 boolean PubSubClient::loop() {
     if (connected()) {
         unsigned long t = millis();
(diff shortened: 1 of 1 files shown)


damdo:
Hey @knolleary would you be able to review/merge this oneliner bugfix?
It would save A LOT of reconnections for all of us!
Thanks :)

hmueller01:
Unfortunately there are so may undone pull requests here that one can believe that @knolleary dropped support of this repo. Maybe it's good to do a PR at https://github.com/thingsboard/pubsubclient as they promised to support development on this library.

damdo:
@hmueller01 opened a PR there: https://github.com/thingsboard/pubsubclient/pull/7

hmueller01:
This is solved in https://github.com/hmueller01/pubsubclient3

PR #1041, opened 2024-02-01
Update PubSubClient.h

anushavenkatesh1:
Increased the buffer size of MQTT_MAX_PACKET_SIZE from 256 to 500


hmueller01:
Why should we increase the default packet size? If you need a bigger buffer use a compiler define `-D MQTT_MAX_PACKET_SIZE 500` or use `PubSubClient::setBufferSize(500)` within your code.

PR #1044, opened 2024-03-05
Fixed a compiler warning.

jstbyte:
Fixed a compiler warning caused by compering unsigned int with signed int.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -520,7 +520,7 @@ boolean PubSubClient::publish_P(const char* topic, const uint8_t* payload, unsig
 
     expectedLength = 1 + llen + 2 + tlen + plength;
 
-    return (rc == expectedLength);
+    return (rc == (unsigned int)expectedLength);
 }
 
 boolean PubSubClient::beginPublish(const char* topic, unsigned int plength, boolean retained) {


hmueller01:
This is duplicate to #1031 ...

hmueller01:
Too bad that @knolleary doesn't maintain this any more. I was hoping the guy's from @thingsboard will do. But they actually don't.

knolleary:
@hmueller01 I'm sorry this has fallen off my radar. I don't actively use it these days, which makes testing and verifying changes tricky.

That said, I'm sure there must be a set of minimal changes just waiting to be sorted - they just need picking out of the backlog of PRs.

If you had any suggestions for what should be prioritised, I'd happily take the input and try to get a new release done with them in next week.

I will then see what I can do to make it clearer in this repo what level of support/help I'm able to provide.

hmueller01:
@knolleary  Nice hearing from you! I‘ll take a look into this tonight and drop a comment on the relevant PRs.

hmueller01:
PS: I can do the testing on ESP8266. I don't have an ESP32 or other boards. So this must be done by someone else.

knolleary:
@hmueller01 I've opened https://github.com/knolleary/pubsubclient/issues/1045 to pull together a plan of what needs to happen in the short term. If you can comment on there with any PRs you think need looking at, that would be super helpful.

bartex:
Once the new release is out, we can test it on ESP32-S3.

PR #1050, opened 2024-04-03
fix: connect failed - no CPU time for IP stack

devmirek:
ESP32-C3 is not able to receive CONNACK packet because the waiting loop did not have any delay to let other tasks to process incoming TCP packet data.

### DIFF
--- a/src/PubSubClient.cpp
+++ b/src/PubSubClient.cpp
@@ -255,6 +255,7 @@ boolean PubSubClient::connect(const char *id, const char *user, const char *pass
             lastInActivity = lastOutActivity = millis();
 
             while (!_client->available()) {
+                delay(10);
                 unsigned long t = millis();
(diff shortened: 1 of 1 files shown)


EmbeddedDevver:
Better use yield().
Also same as #567  and #901

devmirek:
> Better use yield(). Also same as #567 and #901

I would prefer yield() as well. I have tested it with yield() first. It didn't work. Small delay was the only option that reliably solved it (ESP32-C3 with WiFiClientSecure). It was a very tricky issue because sometimes it was successfully reconnected (CONNACK was received after reconnection) but the first connect always failed and sometimes the next reconnects randomly failed as well - always due to CONNACK timeout.

hmueller01:
@devmirek Can you reproduce the problem with https://github.com/hmueller01/pubsubclient3 ?

PR #1064, opened 2024-08-14
Fixed max value of "Remaining length" in "Fixed header"

Nezgar:
According to MQTT 3.1 Specification number of bytes in "Control Packet" may be from 1 to 4(up to 256MB). When we try to send data more than 65535(2 bytes) we overflow value in "length" variable. We cant send data more than 65535(2 bytes).


hmueller01:
I'll soon release PubSubClient3 v3.1.0 that includes sending big payloads.
It is already integrated in the head of
using consequently size_t and can be tested ...

Issue #1045, opened 2024-03-08
Future of this module


knolleary:
I haven't been paying this code the attention it deserves. The issues and PRs have built up to the point where there are lots of duplicates and stale issues.

The code is pretty much in maintenance mode - I'm not looking to add new features etc. But if there are bugs, we should get them fixed.

My rough plan is:

1. Identify a minimal set of PRs needed to address the most common issues - compiler warnings etc
2. Close off all 'old' issues and PRs
3. Publish a new version of the module
4. Update the README to reflect the general 'maintenance' mode of this module

dominicdumont:
Just wanted to mention that this is the best library around for pub/sub. It is very useful and I am grateful that you are still willing to maintain it. Thanks for the great work!

hmueller01:
I scanned all the PR's and would approve to integrate the following fixes:
1. All the same, would take the first/oldest: #788, #826, #899, #1005 or #1031
2. **IMPORTANT** #783 - closes #747, #1020, #745
3. #1038
4. #1029
5. #1001 - if @feedehc accepts my suggestion (or you do it)
6. #993, #863 can be closed with this as well
7. #844
8. #605
9. #567 - closes #901, #314
10. #1046, closes #840

This is potentially worth checking for integration (not adding new features):
1. #1026
2. #966
4. #835 
5. #830 
6. #768

ekazakas:
would it make sense to create an organization and assign several maintainers instead? This is the best MQTT client for Arduino and ESP as far as I know, would be a shame to slowly put it to rest

hmueller01:
I merged almost all patches above, except #1029, #1026 and #835 into
https://github.com/hmueller01/pubsubclient/tree/dev-fixes
If you like you can test it. I tested this on ESP8266 and found no issues til now.
Not sure if #768 works on all platforms, as it uses printf which is not defined in [default Arduino](https://www.arduino.cc/reference/en/language/functions/communication/print/)

cchaz003:
@hmueller01 I'm trying out your changes on some of my deployed devices (the non-critical ones at least). I'll let you know if I run into any bugs related to the merged updates. 

@kamiKAZIK I too would love to see some form of a community support moving forward if @knolleary is on board. I've used this lib under the hood on essentially every MQTT based project I've done over the years and would love to give back. I would be happy to join with a group of like minded folks who want this library to live on and continue developing it!

abdosn:
[...]

PR #1075, opened 2025-02-19
fix: resolve sign comparison warning in PubSubClient::publish_P

SebDominguez:
Changed the type of `expectedLength` from `int` to `unsigned int` to match `rc` type,
resolving the -Wsign-compare warning on line 523: return (rc == expectedLength);


hmueller01:
Welcome to the long list of users fixing this ... ;-)
I think this won't be merged here. See #1045 
You may use my fork, fixing this and lot's of other stuff.
PubSubClient3 should be visible in Arduino and PlatformIO as well.

SebDominguez:
Thanks for maintaining this library. Unfortunately, PubSubClient3 fail to build when using the SAMD21G18A platform, so I just patched version 2.8 for now.

hmueller01:
Can you create an issue at PubSubClient3 and explain what does not work? Do you use 
`platform = SAMD21G18A`
I can't find that at PlatformIO ...

SebDominguez:
Sure, I'm opening the issue right now. 

Issue: https://github.com/hmueller01/pubsubclient3/issues/6

